PR #4, opened 2020-04-14
Fix a null pointer exception on non-cutout devices

fa-gray:
It's a simple enough change, and after this it no longer errors out for my non-notched device.


MathewSachin:
Thank you! I had only tested on Android 9 devices with notches.
So, didn't know that could be `null`.

PR #5, opened 2020-04-14
Add foreground service permission

fa-gray:
This permission is required for some android sdk versions for foreground
notifcations to work.


MathewSachin:
Didn't knew this permission was required starting at Android 9. My bad at having an Android 7.1 phone. Thank you for fixing this!

PR #11, opened 2020-04-17
Stop MediaProjection when stopping the Service

### LINKED ISSUE #10: MediaProjection shows up red after stopping the service
After running the app, there's a red MediaProjection icon that doesn't disappear.

Only force quitting the app removes the icon, which breaks the accessibility setting.

reconman:
Fixes #10

### DIFF
--- a/FateGrandAutomata/ScriptRunnerService.cs
+++ b/FateGrandAutomata/ScriptRunnerService.cs
@@ -90,6 +90,7 @@ namespace FateGrandAutomata
             _virtualDisplay = null;
 
             _mediaProjection?.Stop();
+            _mediaProjection = null;
 
             _windowManager.RemoveView(_layout);
             ServiceStarted = false;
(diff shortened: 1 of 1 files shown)


FateGrandAutomata/ScriptRunnerService.cs
```diff
@@ -89,6 +89,8 @@ public bool Stop()
             _virtualDisplay?.Release();
             _virtualDisplay = null;
 
+            _mediaProjection?.Stop();
+
```
MathewSachin, line 93:
```suggestion
            _mediaProjection = null;

```

We'd also need to set the MediaProjection to `null` if we're doing this.
Otherwise, If the user clicks on `Toggle Service` again after stopping the service, the app will crash. We need to ask the user for the Media Projection permission again.
I've checked that AnkuLua also works in a similar way by asking the permission every time you start the service.

Issue #12, opened 2020-04-17
Replace card priority String with a drag and drop Listview


reconman:
In the Ankulua script, there are 2 options for card priority, simple and advanced mode.

Simple only offers setting the priority of Buster, Quick, and Arts. Advanced offers setting the priority of Weak Buster, Normal Buster, Resistant Buster, ...

I think the App should make it possible to set the priority using a Listview where the entries can be dragged and dropped. The higher up they are, the higher the priority.

I have already found a few projects for Listviews with drag and drop functionality, for example https://help.syncfusion.com/xamarin/listview/item-drag-and-drop

Btw, I personally always use "WB, WA, WQ, B, A, Q, RB, RA, RQ". We could use that as default priority.

MathewSachin:
This was something I was planning on doing eventually anyway.
I think I might be able to do this without requiring additional libraries.

MathewSachin:
I've implemented it, though not much tested yet. I've linked the PR to this issue.

MathewSachin:
@reconman I'm also planning to use sliders in place of numeric textboxes for `Support_SwipesPerUpdate` and `Support_MaxUpdates`. I'll set the minimum to 0 and use defaults of 10 and 3 respectively like in the original script.

What do you think I should keep for their maximum values?

reconman:
For `SwipesPerUpdate`, on my device it needs 9 swipes to get to the bottom when I've disabled all filters. I guess 15 would be a possible max value?

 For `Support_MaxUpdates` I often want to tell the script to update infinitely until I find a support with an MLB CE. Now infinite can be a bad option because if something goes wrong, the script will just run forever.

If I remember correctly, you always need to wait for 10 seconds before you can refresh again. I guess 5 minutes of automatic searching would be a good limit where it doesn't just let the script run infinitely but also allows users to set their script to be more desperate on getting MLB support CEs.

5 minutes = 300 seconds = 30 updates

MathewSachin:
I found that JP has something like an ALL tab on which all servants are shown together. This can be a really long list. I'm not sure if servants other than your friends are also shown in that list though.

reconman:
> I found that JP has something like an ALL tab on which all servants are shown together.

PR #13, opened 2020-04-17
Card priority drag list

MathewSachin:
Requested in #12 

Screenshot_2020-04-17-17-44-43-398_com mathewsachin8 fategrandautomata


reconman:
A couple of notes:
1. You can drag the entries outside of the visible area. If you drag it too high, it will disappear behind "Card types on Top". If you drag it too low, it will disappear behind the white rectangle at the bottom.<br/><br/>Maybe there is a way to limit the drag area?
2. I cannot distinguish the colors of Resist/Normal/Weak, there's barely a difference.

Other than that good work!

MathewSachin:
I've tweaked the colors a bit.

Made it so that the items can only be sorted by dragging the hamburger icons on the right.
Otherwise, we won't be able to scroll the list in Landscape view.

Also, made the list take up the bottom space. So, the white rectangle at the bottom problem is resolved a bit.

MathewSachin:
I think this is in a good enough state to merge

Issue #14, opened 2020-04-17
Make the play button movable like in Ankulua


reconman:
When I have started the service and want to change a setting in the app, the play button will hover over the button for switching the keyboard from ABC to special characters.

I have to stop the service to make the play button disappear.

There should be a way to drag and drop the play button like in Ankulua.

PR #15, opened 2020-04-17
Fix AccessibilityService on force quitting

### LINKED ISSUE #3: Run AccessibilityService in different process
Right now, if you kill the `MainActivity` by swiping the app from task list, then the process is killed.
The `AccessibilityService` running in the same process also gets killed.
But, the switch in the `Accessibility settings` still shows it as `ON`.
Next time you launch the app, you have to go to `Accessibility settings`, turn `OFF` the service and then turn it `ON` again.

The tradeoff with this bad experience is that it is simple to implement.
A `static` instance of the `AccessibilityService` can be used for communication.

This problem can be solved if we run the `AccessibilityService` in a different process.
But, communication with it will be a challenging task here. `Messenger IPC` looks the most promising here. There is also the problem with `SharedPreferences` not working across processes.

---

## I tried implementing this, but multiple processes were not working well. 
I'm dropping this for now.

reconman:
Fixes #3 

I found a way to not break the AccessibilityService when force-quitting the app. The Service (including the play button) will keep running no matter how often you close the app via the task manager.

That's also how Ankulua behaves.

### DIFF
--- a/FateGrandAutomata/ScriptRunnerService.cs
+++ b/FateGrandAutomata/ScriptRunnerService.cs
@@ -167,19 +167,18 @@ namespace FateGrandAutomata
             ShowStatusNotification("Ready");
         }
 
-
-        public override void OnTaskRemoved(Intent rootIntent)
+        public override void OnTaskRemoved(Intent RootIntent)
         {
             // from https://stackoverflow.com/a/43310945/5971497
 
-            Intent restartServiceIntent = new Intent(this.ApplicationContext, this.Class);
-            restartServiceIntent.SetPackage(this.PackageName);
+            var restartServiceIntent = new Intent(ApplicationContext, Class);
+            restartServiceIntent.SetPackage(PackageName);
 
-            PendingIntent restartServicePendingIntent = PendingIntent.GetService(this.ApplicationContext, 1, restartServiceIntent, PendingIntentFlags.OneShot);
-            AlarmManager alarmService = (AlarmManager)ApplicationContext.GetSystemService(Context.AlarmService);
+            var restartServicePendingIntent = PendingIntent.GetService(ApplicationContext, 1, restartServiceIntent, PendingIntentFlags.OneShot);
+            var alarmService = (AlarmManager)ApplicationContext.GetSystemService(AlarmService);
             alarmService.Set(AlarmType.ElapsedRealtime, SystemClock.ElapsedRealtime() + 1000, restartServicePendingIntent);
 
-            base.OnTaskRemoved(rootIntent);
+            base.OnTaskRemoved(RootIntent);
         }
 
         Button _scriptCtrlBtn;
--- a/FateGrandAutomata/ScriptRunnerService.cs
+++ b/FateGrandAutomata/ScriptRunnerService.cs
@@ -18,7 +18,7 @@ using Java.Interop;
 namespace FateGrandAutomata
 {
     [Service(Permission = Manifest.Permission.BindAccessibilityService)]
-    [IntentFilter(new [] { "android.accessibilityservice.AccessibilityService" })]
+    [IntentFilter(new[] { "android.accessibilityservice.AccessibilityService" })]
     [MetaData("android.accessibilityservice", Resource = "@xml/script_runner_service")]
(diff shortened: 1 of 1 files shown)


MathewSachin:
THIS is really great!
I was trying to do this in like forever, to no success.
I'm right now working on #14, I've almost found a solution.
Once I open a PR for that, I'll surely test this out.
Big Thanks!

MathewSachin:
My phone is a Xiaomi one, so this doesn't work by default on it.
I've to go to the phone's settings and enable an option called `Autostart`.
After that, works as expected.

See https://stackoverflow.com/questions/40660216/ontaskremoved-not-getting-called-in-huawei-and-xiomi-devices

I checked and found that Autostart was NOT enabled for AnkuLua.
So, they might be using some other way to restart the AccessibilityService.

reconman:
So for now we should add the code in the answer, right?

reconman:
Btw, you are able to add commits to my branch.

MathewSachin:
> So for now we should add the code in the answer, right?

I think we probably should not. That answer is for Xiaomi devices only. Other answers suggest that this problem can happen on other phones too. And the setting to be enabled can vary depending on the device.
I think information about this should go into the README.

> Btw, you are able to add commits to my branch.

I should be by default, there should be an `Allow edits from maintainers` checkbox on the PR sidebar which should be checked by default.

FateGrandAutomata/ScriptRunnerService.cs
```diff
@@ -167,6 +167,21 @@ void StopScript()
...
+
+            Intent restartServiceIntent = new Intent(this.ApplicationContext, this.Class);
+            restartServiceIntent.SetPackage(this.PackageName);
+
+            PendingIntent restartServicePendingIntent = PendingIntent.GetService(this.ApplicationContext, 1, restartServiceIntent, PendingIntentFlags.OneShot);
+            AlarmManager alarmService = (AlarmManager)ApplicationContext.GetSystemService(Context.AlarmService);
+            alarmService.Set(AlarmType.ElapsedRealtime, SystemClock.ElapsedRealtime() + 1000, restartServicePendingIntent);
+
+            base.OnTaskRemoved(rootIntent);
+        }
```
MathewSachin, line 183:
```suggestion
        public override void OnTaskRemoved(Intent RootIntent)
        {
            // from https://stackoverflow.com/a/43310945/5971497

            var restartServiceIntent = new Intent(ApplicationContext, Class);
            restartServiceIntent.SetPackage(PackageName);

            var restartServicePendingIntent = PendingIntent.GetService(ApplicationContext, 1, restartServiceIntent, PendingIntentFlags.OneShot);
            var alarmService = (AlarmManager)ApplicationContext.GetSystemService(AlarmService);
            alarmService.Set(AlarmType.ElapsedRealtime, SystemClock.ElapsedRealtime() + 1000, restartServicePendingIntent);

            base.OnTaskRemoved(RootIntent);
        }
```

Just a few conventions I follow.
I'm just trying whether I can commit to your branch since you asked.

MathewSachin:
> Btw, you are able to add commits to my branch.

Yup, I checked it. I made minor code style changes.

reconman:
> I should be by default, there should be an Allow edits from maintainers checkbox on the PR sidebar which should be checked by default.

Just wanted to make sure you've seen it.

Also, regarding Ankulua: Are you even using Accessibility mode for clicking? I personally use root for everything.

MathewSachin:
> Also, regarding Ankulua: Are you even using Accessibility mode for clicking? I personally use root for everything.

`AccessibilityService` is the only way if you want to do things like clicks without root. Also, I'm afraid my phone isn't rooted.
F/GO doesn't work on rooted phones by default and I know there's some software called `MagiskHide` or something which can hide the root status from F/GO.
The average user would still not want to root their phone. So, by sticking with `AccessibilityService`, the solution is useful for more people.

Other than this specific problem, I don't have any other complaints with using `AccessibilityService`. Even now, I'm not too bothered, even though I have to turn OFF and ON the service every time I open the app.
I won't have to do that again with this PR.
Thanks again!

PR #16, opened 2020-04-17
Movable Play button

### LINKED ISSUE #14: Make the play button movable like in Ankulua

MathewSachin:
fixes #14 

~~Currently, it is aligned to top-left by default.
I prefer to make it bottom-left by default.~~

### DIFF
--- a/CoreAutomata/ComparableExtensions.cs
+++ b/CoreAutomata/ComparableExtensions.cs
@@ -1,0 +1,18 @@
+﻿using System;
+
+namespace CoreAutomata
+{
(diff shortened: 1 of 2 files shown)


MathewSachin:
@reconman I've implemented this and it's working quite good for me.
You might want to check it out.

Issue #17, opened 2020-04-18
Set up Azure Pipeline for the master branch


reconman:
When anything is committed or merged to the branch, a pipeline should be triggered which builds the current APK from that master branch.

At a later date we could also automate releases, Azure supports publishing to Github.

I've spent quite some time today finding out how to get Azure Pipelines to work with the project. https://dev.azure.com/reconman/Fate-Grand-Automata/_build/results?buildId=2&view=artifacts&type=publishedArtifacts

I would suggest us chatting privately because it involves many steps. I have Discord and Skype accounts.

MathewSachin:
Azure pipelines looks like the only CI option to work out of the box with Xamarin.
I was also thinking about trying out GitHub Actions, maybe later.

Discord: mathewsachin#9362

reconman:
I've sent you a friend request

PR #19, opened 2020-04-18
GitHub Actions CI workflow

MathewSachin:
This will build and package the app as APK on every commit.
The app is securely signed with Release certificate.
Also, creates a GitHub Releases draft with the APK uploaded when a tag is pushed.

~~Need to add versioning.~~

PS: GitHub Actions is really awesome.
It's Lightning fast. I can't imagine my system doing the whole process in 1.5 min.

---

@reconman #17 ~~I think we can even deploy to GitHub Releases from here only.~~


MathewSachin:
GitHub Release drafts will now be automatically created when a tag is created!

MathewSachin:
I think Versioning can be handled manually.

When we want to do a release:
1. Update version code in `AndroidManifest.xml`.
2. Create and push a new tag corresponding to the version
3. GitHub Actions builds the APK and uploads it into a new GitHub Releases draft.
4. We can update the description of the release and publish it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - '*'
```
reconman, line 7: Should we limit it to only master? Otherwise you'll have 2 builds for every PR you create.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      - 'v*'
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Android
+      env:
+        GPG_KEYSTORE: ${{ secrets.GPG_KEYSTORE }}
```
reconman, line 18: You're correctly using secrets here, but you've pushed your keystore in this PR. You should at least replace this commit with another one without the keystore.

But also anyone who can find the old commit hash will be able to download the leaked keystore file.

MathewSachin: keystore isn't leaked. It's GPG encrypted and the key is stored in secrets.
This practice is actually present in the GitHub documentation: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#limits-for-secrets

.github/workflows/ci.yml
MathewSachin: Right, I'll limit it to master

.github/workflows/ci.yml
reconman: Ah, I've never used GPG encryption for anything before, I've only heard about it in the context of encrypting mails.

reconman: But also, do you then still need to have it committed?

MathewSachin: Yeah, GitHub says to do so in their official documentation, so it should be pretty safe. See point 4 in the link above.

reconman: I just noticed, you're using secrets.GPG_KEYSTORE as keystore password and KEYSTORE_PASS as key password.

If they're the same, you can leave out the key password parameter. I would change the secrets like this:
GPG_KEYSTORE -> KEYSTORE_PASS
KEYSTORE_PASS -> KEY_PASS

reconman:
Looks good to me

PR #24, opened 2020-04-19
Fix reading refill repetitions

### LINKED ISSUE #23: Battle can't be started because of ClassCastException on refill repetitions
When you try to start the battle script, it will quit immediately with a ClassCastException

Refill repetitions is stored as a String, but is read like an Integer.

reconman:
Fixes  #23


MathewSachin:
This happened yesterday when I changed the support SwipesPerUpdate and MaxUpdates options to sliders and forgot to update for refill repetitions.
Thank you for fixing!

PR #25, opened 2020-04-19
Fix APK signature

reconman:


reconman:
@MathewSachin Please fix the keystore passwords. https://github.com/MathewSachin/Fate-Grand-Automata/pull/25/checks?check_run_id=598764803#step:5:570

Before, it was not using your keystore but instead used a newly generated one because of a typo.

reconman:
@MathewSachin So because of a typo, the pipeline always used a new keystore for signing.

Additionally, there was a communication error between us. There are up to 3 passwords in this pipeline regarding the keystore:

* Keystore password
  * password for opening the decrypted keystore
  * currently $KEY_PASSWORD
* Key password
  * password for using the key inside the keystore
  * currently $KEY_PASSWORD
* GPG decryption password
  * password for decrypting the keystore
  * currently $KEYSTORE_PASSWORD

I got the pipeline to work again, but I'm not totally happy with the current variable names.

MathewSachin:
I'm sorry, I messed up really bad here.
Thank you for fixing.
I'll change the secret names. Let me know what you want them to be.

reconman:
The names before my review in #19 were better. I just didn't understand the pipeline script well enough.

I would ultimately change them to GPG_PASS (or GPG_KEYSTORE_PASS) and KEYSTORE_PASS

MathewSachin:
I'm changing them back to KEYSTORE_PASS and GPG_KEYSTORE then.
GPG_KEYSTORE sounds better to me = GPG for KEYSTORE and the convention can be followed if we include any other GPG encrypted files.

PR #32, opened 2020-04-20
Fix: Autosupport scrolling too far

### LINKED ISSUE #31: AutoSupport scrolls too far
On my device, the AutoSupport script scrolls too far when no Event Bonus filters are set.

The script will scroll 2 1/2 entries, so it goes from this:

To this:

The amount of scrolling should be reduced, so the 3rd entry is fully shown once.

MathewSachin:
fixes #31 

The scrolling can be controlled precisely if we hold the touch down after the swipe.
But, this feature was added only in Android 8.

So, for Android 7, I'm right now planning a simple fix where the distance to scroll is halved.
This is not exactly accurate but at least no servant/CE will be missed even if the play button is over the support bounds.
I'll see if I can come up with any better way for this.

On Android 8 and above, we can use the `continueStroke` function to do the swipe and hold gesture.

### DIFF
--- a/FateGrandAutomata/AndroidImplGestures.cs
+++ b/FateGrandAutomata/AndroidImplGestures.cs
@@ -38,11 +38,11 @@ namespace FateGrandAutomata
             var holdPath = new Path();
             holdPath.MoveTo(End.X, End.Y);
 
-            swipeStroke.ContinueStroke(holdPath, 0, holdDuration, false);
+            var holdStroke = swipeStroke.ContinueStroke(holdPath, 0, holdDuration, false);
 
             var gestureBuilder = new GestureDescription.Builder();
             gestureBuilder.AddStroke(swipeStroke);
-
+            gestureBuilder.AddStroke(holdStroke);
             PerformGesture(gestureBuilder.Build());
         }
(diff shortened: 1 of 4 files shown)


FateGrandAutomata/AndroidImplGestures.cs
```diff
@@ -0,0 +1,108 @@
...
+            var swipeStroke = new GestureDescription.StrokeDescription(swipePath, 0, duration, true);
+
+            var holdPath = new Path();
+            holdPath.MoveTo(End.X, End.Y);
+
+            swipeStroke.ContinueStroke(holdPath, 0, holdDuration, false);
+
+            var gestureBuilder = new GestureDescription.Builder();
+            gestureBuilder.AddStroke(swipeStroke);
+
```
reconman, line 45:
```suggestion
            var holdStroke = swipeStroke.ContinueStroke(holdPath, 0, holdDuration, false);

            var gestureBuilder = new GestureDescription.Builder();
            gestureBuilder.AddStroke(swipeStroke);
            gestureBuilder.AddStroke(holdStroke);
```

PR #45, opened 2020-04-22
Consistent Default Card Priority

### LINKED ISSUE #43: If the card priority is not set, it will be BAQ, regardless of weak or resistant
After having to reset my app a few times during debugging, I noticed that the app would pick buster cards, even if they were weak.

The only explanation for that is that the default value for the card priority is BAQ instead of wB wA wQ and so on.

A small thing that will probably be fixed in the course of this bugfix: if the user has never opened the card priority menu, the text below it with the card priority String will be empty.

MathewSachin:
fixes #43


reconman:
Please merge master into this one, then I can test it on my tablet

Issue #53, opened 2020-04-22
Stuck in a start quest loop


Mysty-chan:
> MAKE SURE YOU TESTED THE LATEST RELEASE.

> See 'SAMSUNG.md' first, if you're a Samsung user.

**Describe the bug**
Script keeps pressing the screen at the start quest location no matter what screen is visible. Just catches the first support on the list, then changes party around rather than starting quest. If you start script after selecting support, it keeps clicking that same spot. 
Tried clearing app cache, restarting device, switching to manual support selection. Started second time using ver 10, still occurring with ver 10.1

**Screenshots**
![Screenshot_20200422-151717_Video Player](https://user-images.githubusercontent.com/64169016/80034943-69a6c580-84ac-11ea-9327-14ba96b93dce.jpg)

![Screenshot_20200422-122913_XRecorder](https://user-images.githubusercontent.com/64169016/80034679-087ef200-84ac-11ea-8255-c07177c41aa8.jpg)
![Screenshot_20200422-123101_XRecorder](https://user-images.githubusercontent.com/64169016/80034681-09b01f00-84ac-11ea-828e-fb0c29170ab5.jpg)

**Device Info (please fill atleast 'Device model'):**
 - Device model: samsung galaxy tab a 2017, SM-T380 
 - Screen size: [e.g. 1280x720] (not sure where to find this) 
 - Android Version: 9
 - RAM: 2GB

reconman:
Have you tried uninstalling the app and then installing it again? You will lose your settings (no backup function yet) but it may help.

reconman:
Oh you can find out your resolution by searching for "<model> resolution" or by measuring your screenshots. In your case google tells me it's 1920x1200.

Mysty-chan:
I was afraid uninstalling would loose settings. Let me try saving that manually and un/reinstall

Mysty-chan:
Just finished the reinstall of 10.1 and still having the same problem of the start quest loop

Mysty-chan:
I just tried rolling back to ver 9, which had been working minus the funny Samsung timeout bug after ~1.8 quests. Alas, that now has the bug. So formerly working code is now nonfunctional...which suggests that something changed with my tablet? I was outside doing yard work between the last working session and the first loopy session not changing device settings... Know of any background android/Samsung functions that could screw with non play protect approved apps?

reconman:
First of all, I'm unable to reproduce the issue on a Samsung TM-813.

[...]

PR #56, opened 2020-04-23
Potential fix for the quest selection loop

reconman:
Disable the Listener function, always use a new screenshot.


MathewSachin:
Quest selection loop in #53 was due to the recording app used messing up the `MediaProjection`.
So, we don't need to do this change.

PR #59, opened 2020-04-23
Wait for Support list to load

MathewSachin:
Wait for support list to load before starting support selection code.
The `Confirm Support Setup` for the first servant is checked with a timeout of 30s.
This should be enough time for the list to load even with poor network connectivity or laggy device.

Here's the region checked:
ensure_support


MathewSachin:
This will fail with story support

PR #74, opened 2020-04-26
Allow NP usage after executing commands

echoscrip:
Battle.CurrentStage starts at 0, and so if there were any commands in the _commandTable (making count >=1) then IsFinished will never be set to true for the any stage with commands.


MathewSachin:
Yeah, you're right. I missed that due to indexing in Lua starting at 1 and in C# at 0.
Thank you for fixing!

The build failing problem is from our side.
I'll fix the build script and let you know to merge it into your branch.

MathewSachin:
I'll merge it right now! Thank you for the contribution!
We'll try to get the CI script working for external Pull requests soon.

PR #82, opened 2020-04-26
Add all support images from the original repo

potchy:
This PR includes all CEs and Servants from the original repo.
Servants are organized by folder, as per the new structure.

I'm sure the APK size will linearly increase as we add more images, so maybe these should be distributed as a separate package instead?


MathewSachin:
@potchy We're actually planning to keep the images organized in a separate repository.
We're still in the process of deciding file name conventions to avoid duplicates and all.
I'll let you know once the conventions are ready or we could handle adjusting the filenames ourselves if you're too busy and all.

MathewSachin:
I've done it again. I've migrated the whole codebase to Kotlin, PR in review #108.
Xamarin C# was giving us some problems like weird errors on the latest version (Android 7.1.2) of Nox emulator.
Also, memory leaks were a pain to debug with limited support for free users of Visual Studio.

So, this app is now a proper Android app which is developed in Android Studio.
As a result, the directory structure for the project will change heavily causing even more conflicts with this PR. So, I'm closing it.

BTW, thank you for your work on the original Lua project. 👍 
I'll be happy to have you here too if you got time.

PR #84, opened 2020-04-26
Fix CI for external PRs

MathewSachin:
1. Introduced a `fgadebug.keystore`.
2. Changed package name to `com.mathewsachin.fategrandautomata.debug` when using for debug purposes.
3. CI changes package name to `com.mathewsachin.fategrandautomata`.
4. Sign using debug key for forks
5. Fork builds change package name to: `com.mathewsachin.fategrandautomata.fork` to allow side-by-side installs.

So, we have 3 versions of the app that can be installed side-by-side:
1. debug
2. Release/CI
3. fork


.github/workflows/ci.yml
```diff
@@ -11,39 +11,74 @@ on:
...
+    env:
+      NOT_FORK: ${{secrets.NOT_FORK == 'true'}}
+      FORK: ${{secrets.NOT_FORK != 'true'}}
     steps:
     - uses: actions/checkout@v1
+
+    - name: Package Name
+      run: sed -i -e "s/package=\"[^\"]*\"/package=\"com.mathewsachin.fategrandautomata\"/g" AndroidManifest.xml
+      working-directory: ./FateGrandAutomata/Properties
+      if: env.NOT_FORK == 'true'
```
reconman, line 23: I think you can leave out the == 'true', Github Actions can deal with booleans

.github/workflows/ci.yml
```diff
@@ -11,39 +11,74 @@ on:
...
+
+    - name: Package Name (Fork PR)
+      run: sed -i -e "s/package=\"[^\"]*\"/package=\"com.mathewsachin.fategrandautomata.fork\"/g" AndroidManifest.xml
+      working-directory: ./FateGrandAutomata/Properties
+      if: env.FORK == 'true'
```
reconman, line 28: Same here

.github/workflows/ci.yml
```diff
@@ -11,39 +11,74 @@ on:
...
   build:
     runs-on: macos-latest
+    env:
+      NOT_FORK: ${{secrets.NOT_FORK == 'true'}}
+      FORK: ${{secrets.NOT_FORK != 'true'}}
     steps:
     - uses: actions/checkout@v1
+
+    - name: Package Name
+      run: sed -i -e "s/package=\"[^\"]*\"/package=\"com.mathewsachin.fategrandautomata\"/g" AndroidManifest.xml
```
reconman, line 21: How about using a bash if else instead?

I'm not really familiar with bash, but it should be like
```
if [$NOT_FORK == "true"]; then
    PACKAGE_NAME="com.mathewsachin.fategrandautomata"
else
    PACKAGE_NAME="com.mathewsachin.fategrandautomata.fork"
fi
sed -i -e "s/package=\"[^\"]*\"/package=\"$PACKAGE_NAME\"/g" AndroidManifest.xml
```

This will save some copy pasting.

.github/workflows/ci.yml
```diff
@@ -11,39 +11,74 @@ on:
...
+
+    - name: Android
+      env:
         KEYSTORE_PASS: ${{ secrets.KEYSTORE_PASS }}
-      run: |
-        cd FateGrandAutomata
-        gpg --quiet --batch --yes --decrypt --passphrase="$GPG_KEYSTORE" --output fgautomata.keystore fgautomata.keystore.gpg
-        msbuild FateGrandAutomata.csproj /restore /verbosity:minimal /t:SignAndroidPackage /p:Configuration=Release /p:AndroidKeyStore=True /p:AndroidSigningKeyAlias=fgautomata /p:AndroidSigningStorePass="$KEYSTORE_PASS" /p:AndroidSigningKeyPass="$KEYSTORE_PASS" /p:AndroidSigningKeyStore=fgautomata.keystore
+      working-directory: ./FateGrandAutomata
+      if: env.NOT_FORK == 'true'
```
reconman, line 52: bash if else could also be used here

.github/workflows/ci.yml
reconman: PACKAGE_NAME could be shared with the Upload APK step: https://www.edwardthomson.com/blog/github_actions_15_sharing_data_between_steps.html

.github/workflows/ci.yml
```diff
@@ -11,39 +11,74 @@ on:
...
+      run: sed -i -e "s/package=\"[^\"]*\"/package=\"com.mathewsachin.fategrandautomata.fork\"/g" AndroidManifest.xml
+      working-directory: ./FateGrandAutomata/Properties
+      if: env.FORK == 'true'
+
     - name: Version Number
+      working-directory: ./FateGrandAutomata/Properties
       run: |
-        cd FateGrandAutomata/Properties
         sed -i -e "s/android:versionCode=\"[^\"]*\"/android:versionCode=\"$GITHUB_RUN_NUMBER\"/g" AndroidManifest.xml
         sed -i -e "s/android:versionName=\"[^\"]*\"/android:versionName=\"$GITHUB_RUN_NUMBER\"/g" AndroidManifest.xml
```
reconman, line 34: We can also do a bash if else here instead of running 2 steps for tags.

.github/workflows/ci.yml
```diff
@@ -11,39 +11,74 @@ on:
 jobs:
   build:
     runs-on: macos-latest
+    env:
+      NOT_FORK: ${{secrets.NOT_FORK == 'true'}}
+      FORK: ${{secrets.NOT_FORK != 'true'}}
```
reconman, line 16: You can remove FORK, I think, because you should be able to use boolean logic in Github Actions ifs. Like `!NOT_FORK`. But maybe it's then better to use FORK instead? Not not fork is confusing.

.github/workflows/ci.yml
MathewSachin: It doesn't. I did that before and got errors. Environment variables are always treated as strings.

.github/workflows/ci.yml
MathewSachin: `!NOT_FORK` doesn't work. already tried. Environment variables are treated as strings.
And, yeah I added `FORK` to reduce confusion from not of not.

PR #88, opened 2020-04-27
Bigger btn on bigger screens

### LINKED ISSUE #61: Make the play/stop button a bit bigger
I have trouble getting the button to activate sometimes. A little bigger, or option to adjust the size per user's preference, would be nice. No rush though, it's a little thing.

MathewSachin:
fixes #61


FateGrandAutomata/ScriptRunnerService.cs
```diff
@@ -284,7 +283,9 @@ void ScriptCtrlBtnOnTouch(object S, View.TouchEventArgs E)
 
                     var d = Math.Sqrt(Math.Pow(newX - _layoutParams.X, 2) + Math.Pow(newY - _layoutParams.Y, 2));
 
-                    if (_dragging || d > DragThreshold)
+                    var dragThreshold = _layout.MeasuredWidth;
```
reconman, line 286: So you're checking if the user drags it for the width of the button? Because there's also https://developer.android.com/reference/android/view/ViewConfiguration#getScaledTouchSlop()

MathewSachin: That value is 16 is on my phone.
When I was using DragThreshold value of 20-30, it was too sensitive to dragging.
I think `ScaledTouchSlop` might be more related to scrolling than dragging.

Using 70-100 was when the sensitivity became right.
I changed it to use the width of the button because bigger buttons would otherwise drag easily since the user has more space to press.

MathewSachin: Otherwise, we can make it long press to drag.

MathewSachin: Time-based dragging instead of distance-based dragging is actually working fine!

PR #90, opened 2020-04-28
ScreenShot using Root

MathewSachin:
The app will now work with NoxPlayer Android 7 for FGO JP on which Media Projection doesn't work otherwise.


FateGrandAutomata/SuperUser.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public void SendCommand(string Command)
+        {
+            _superUserStreamWriter.WriteLine(Command);
+            _superUserStreamWriter.Flush();
+
+            // Wait
+            // https://stackoverflow.com/a/16160785/5377194
+            {
```
reconman, line 34: Any reason for `{`?

MathewSachin: No, it was just there for attribution as a block. I'll add it to separate function.
I'm refactoring the code a bit and adding the root gestures. Will be done in a few minutes.

PR #94, opened 2020-04-30
Support Image Renamer

MathewSachin:
A window pops up after running Support Image Maker script where you can rename/delete the files.

Some manufacturers prevent services from opening Activities. For example, on Xiaomi phones, you need to enable `Display pop-up window while running in the background` in `App Info / Other Permissions` every time the app is installed.
But, even if you don't enable the setting, Support Image Maker continues to work as it did before.


MathewSachin:
Just write the servant's name as `Arjuna Alter/asc1`, `Arjuna Alter/asc2`.
I had made sure this works.

reconman:
Okay, that's definitely not intuitive, so users would have to read the documentation, which is not really good.

MathewSachin:
Yeah, I was also thinking of the dropdown before but costumes turned out to be the problem

reconman:
I'm getting an error "Failed to rename to: Mordred/asc2". I expected some code to create the folder if it doesn't exist.

MathewSachin:
My bad. I'll fix that

MathewSachin:
Added code for creating folder if it does not exist

reconman:
One more issue, you can do that tomorrow:

When you click on the green checkbox without specifying one or more names, it will keep the files with the old names. I would personally not do that. I would show a message that the user should either delete the entries or give them a name.

MathewSachin:
That was why I was showing the original names, but, yeah, your idea is better.
But, what about user pressing the back button or closing the activity? The files would have the same name then also.

BTW, is that check icon showing as Green on your phone? It should've been white

reconman:
We could make the back button either:
* Delete the files
* Leave the files

I think it would be more intuitive for normal users if we delete the files on pressing the back button. How do you close the activity without pressing the back button?

The check button was indeed white.

And regarding the initial user request of being able to delete files after an event is over, it could be possible to open a file explorer app to the location of the folder: https://stackoverflow.com/a/26651913/5971497

But that only works if the user has an app for that.

MathewSachin:
We should at least show a prompt before deleting the files.
The users can also close the activity by swiping from the task list.
I'm against blindly deleting the files because the Support Image Namer might not even show up for Xiaomi users like I wrote in the first comment. Might be the case with some other manufactures too.

And, I'd seen about opening file manager, but there is no standard followed by all file managers.
So, I don't think we should do that.

reconman:
There are 3 issues for me:
1. You can click the checkbox without specifying names
2. When the support preferences screen is open and files change, the lists won't be updated
3. You are currently allowed to specify folder names for CEs, but folders are not handled by the CE image list code. We should disallow folders in CE names.

reconman:
Sorry for the late feedback.

This is how the error message looks like right now:
image

I would replace `\n` with `\n\n` to make it easier to read. The last sentence of `Avoid unnecessary spaces` is not really clear. People will not use spaces in their names if you write it like that.

FateGrandAutomata/SupportImgNamerActivity.cs
```diff
@@ -199,10 +199,10 @@ bool Validate(string OldPath, string NewFileName, Regex Regex, string InvalidMsg
         const string InvalidCharsMsg = "'<, >, \", |, :, *, ?, \\, /'";
```
reconman, line 199: Last thing, I would remove the `'` in this message. Otherwise people might think that `'` is included in the character list

Ansem93:
Sorry to bump this ticket. Where can I find the list of CE and servants? I can't find anything in "more option" where i could rename CE and servants

reconman:
@Ansem93 Doesn't exist anymore.

Rename the files in your support directory using a file explorer.

You selected the folder during the initial configuration.

PR #95, opened 2020-05-01
Implement Debug mode using Region.Highlight

MathewSachin:


CoreAutomata/TransformationExtensions.cs
```diff
@@ -35,15 +35,24 @@ public static Location Transform(this Location Location)
         {
             var scale = ScriptToScreenScale();
 
-            var scaledPoint = scale == null
-                ? Location
-                : Location * scale.Value;
+            var scaledPoint = Location * scale ?? Location;
```
reconman, line 38: This expression is written in a confusing way for non C# developers.

I personally would prefer `= Lociation * (scale ?? 1)`. Or better yet, why not return 1 in ScriptToScreenScale()?

FateGrandAutomata/AndroidImpl.cs
```diff
@@ -55,6 +62,14 @@ public void MessageBox(string Title, string Message)
             });
         }
 
+        public void Highlight(Region Region, double Timeout)
+        {
+            HighlightView.AddRegion(Region);
+
+            Task.Delay(TimeSpan.FromSeconds(Timeout))
```
reconman, line 69: This looks like an asynchronous call, won't the highlights disturb the screenshots then?

So I would add an `await` here.

FateGrandAutomata/AndroidImpl.cs
```diff
@@ -55,6 +62,14 @@ public void MessageBox(string Title, string Message)
             });
         }
 
+        public void Highlight(Region Region, double Timeout)
```
reconman, line 65: How about replacing double with TimeUnit everywhere? That way you won't make the mistake again of not knowing what time unit it is.

CoreAutomata/AutomataApi.cs
```diff
@@ -180,6 +192,11 @@ public static void ShowMessageBox(string Title, string Message)
             _platformImpl.MessageBox(Title, Message);
         }
 
+        public static void Highlight(Region Region, double Timeout)
+        {
+            _platformImpl.Highlight(Region.Transform(), Timeout);
```
reconman, line 197: All the other ones use Region.Highlight() but this one uses _platformImpl.Highlight() with a transformed region?

MathewSachin:
Since most of my responses have some connection with each other, I'm writing a common one first.
This became such a big comment. 😆 

### What I wanted debug mode to be
I wanted debug mode to highlight the regions as well as keep the script running at the same speed.
Otherwise, what's really the use if the script will just wait after each highlight. Waiting will in many cases change the script's behaviour. We want it to run in the exact same way as it does without debug mode so that we can see what the script was doing at some point. And, I've confirmed multiple times that this DOES NOT interfere with the screenshots enough to cause any change in its behaviour. It doesn't matter in most cases since we use the `UseSameSnap` for many things which uses the same screenshot for further matching without taking new ones.
The script only highlights a region for 300ms right now. This can surely be increased.

### Highlighted regions are too far to the right
Can you give me a screenshot for this? I tested both on my phone and Nox and the regions seem fine to me. Yeah, if you are talking about the rectangles going out of the screen, when they are at screen's border, I actually thought it would be better that way, so that future screenshots won't be overlapped. Might be having some problems on non-16:9 layouts.

### About writing Comments
I admit I'm not very good at documentation and writing comments. The comments I write usually ends up feeling redundant to me, so I mostly give up on them. I really should improve myself here.
Instead of saying that the changes in this PR were difficult to understand, I think the overall project structure is a bit difficult to understand. I've explained it a bit below.

### Returning `null` in scale functions
It was just an optimization to avoid unnecessary scaling of Images. You might then say then why not check `scale== 1`, but there is a problem, which is that equality comparison with floating-point types is not precise. Instead of `scale == 1` we should ideally do `Math.Abs(scale - 1) < epsilon` where `epsilon` is the significance we want. I learned this from `Resharper`. It complains every time I compare two `double` or `float` values. This won't really be a problem usually but is not a good thing to do.
But, yeah, I agree that writing it as `Location * (scale ?? 1)` would be better.

### How `AutomataApi` works
`AutomataApi` is just a proxy for calling the actual methods which are registered by an `IPlatformImpl`. `AndroidImpl` is where the actual `Highlight` function is defined. `Region.Highlight` is just a convenience function which itself calls `AutomataApi.Highlight`.

Now, the code in `FateGrandAutomata.Core` project works in the `ScriptDimension` coordinate space. And, the Android project works in the real coordinate space. It is the job of `AutomataApi` to transform the `Location`, `Region`, etc. objects to the correct coordinate spaces. This keeps those projects unaware of the transformations happening. It also handles changing images and corresponding `Region` to the `CompareDimensions` coordinate space.

I wrote it this way so that porting the scripts from Lua is easy. The scripts continue to work in the same dimensions they did before.

### Time Unit
The whole Lua code was written using decimal times, so for ease of porting, I did the same. A simple regex replace should work for changing to `TimeSpan` objects, but that should be done in a different PR. The problems I had can happen even when using `TimeSpan` objects when we're interacting with Android's APIs though.

reconman:
Here are my reactions:

## What I wanted debug mode to be

The real problem with not waiting is that you can never take screenshots of them as a user because the red lines are gone too quickly. You can only detect problems if you do a video recording and look at it closely.

And if just decide to expand their lifetime from 0.3 to say 0.6, then suddenly you'll have 10 rectangles on the screen at the same time.

## Highlighted regions are too far to the right
Here you can see that the red rectangle is off to the right.



The offset seems to be the width of my notch.

## About writing Comments

I get the overall project structure and the connections, but still people who just joined developing this app would have to research every Android and OpenCV method you're using to find out what's happening.

In my old Github project almost nothing was commented and nobody except for me knew what things did. The useful things about method comments is that they appear when you hover over the methods in Visual Studio.

Comments may feel redundant to you because you currently know what the code does but this is not about you understanding your own code, it's about making it easier for other developers.

## Returning null in scale functions

Is the value returned by `ScriptToScreenScale()` currently being compared anywhere to make your float comparison argument relevant?

## How AutomataApi works

But then why does AutomataApi call `Region.Highlight()` instead of `this.Highlight()`? It currently goes like this: AutomataApi.FindAll() -> Region.Highlight() -> AutomataApi.Highlight() -> AndroidImpl.Highlight()

## Time Unit

Yes, the danger of using the wrong time unit will also be there in the future, but at least it's clear for any developer working on your project. You called your parameter Timeout and expect readers to be able to detect that `double` implies seconds.

If you use `TimeSpan.FromSeconds()` or `TimeSpan.FromMilliseconds()` and then use that object somewhere else, you can just use `timeSpan.TotalSeconds` and `timeSpan.TotalMilliseconds` instead of relying on the datatype or variable to tell you which one it is supposed to be.

I only commented about using TimeSpan because you were adding a new method and mentioned yourself how we could use that class instead of double or int. So I thought "Why not do it cleanly for new methods?".

MathewSachin:
Reactions of reactions:

### Highlighted regions are too far to the right
It seems we can't draw over the notch area, I'll adjust the drawn rectangles while also considering the Samsung option, in which case we can draw over the notch area.

### Returning null in scale functions
Yeah, `ScriptToScreenScale` is used only for `Region`, `Location`, etc.
Only `ScreenToImageScale` is used to resize image. Sure, I'll make it return `1` instead of `null` then.

### Why does AutomataApi call Region.Highlight() instead of Highlight(Region)
Because it looks better 😄 

### Time Unit
Sure, I'll make the highlight code use `TimeSpan` objects.

### Comments
You know that's the real problem when you are the code author. You don't know exactly what needs to be explained and how much unless you're an experienced developer who has guided some assistants.
Still, I'll try to write some comments in a separate PR.

### What I wanted debug mode to be
I think we need to discuss this more. If we only want the screenshot of a region from user, why not make it the whole game area, other areas would be relative to it anyway.
I actually found debug mode working this way useful in that now I know exactly when the script is doing what on my phone. I can see when it's in the outer loop, when it's in support, etc.
I can also see when it tries to check for Mana Prism/SQ rewards, which almost always misses on my phone.

FateGrandAutomata/AndroidImpl.cs
```diff
@@ -66,6 +66,11 @@ public void Highlight(Region Region, TimeSpan Timeout)
         {
             HighlightView.AddRegion(Region);
 
+            // We can't draw over the notch area
+            var cutoutAppliedRegion = CutoutManager.GetCutoutAppliedRegion(_accessibilityService);
+            Region.X -= cutoutAppliedRegion.X;
```
reconman, line 71: Don't you need to clone the Region first?

MathewSachin: It's a region generated from transform, so, it's already cloned.

I've kept `Region` as a `class` when ideally it should've been a `struct`, so this works, but, I should do this before `AddRegion` for clarity.

CoreAutomata/TransformationExtensions.cs
```diff
@@ -79,21 +74,18 @@ public static Region TransformFromImage(this Region Region)
...
 
-        static double? ScriptToScreenScale()
+        const double DontScale = 1;
+
+        static double ScriptToScreenScale()
         {
             if (GameAreaManager.ScriptDimension == null)
             {
-                return null;
+                return DontScale;
```
reconman, line 88: Why don't you also do this for ScreenToImageScale()?

MathewSachin: `ScreenToImageScale` is used by `ScreenshotManager` to decide if the screenshot needs to be resized. if `null`, don't resize. I don't want to check `scale == DontScale` due to the issue I wrote about in floating-point equality check.

MathewSachin: Oh, I think I get what you're saying now.
You meant `scale ?? DontScale` instead of `scale ?? 1` right? I'll make the change.

PR #100, opened 2020-05-03
Autoskill Maker

MathewSachin:
To make it easier for people to make Autoskill commands.

Still in pretty early stages.
Can be opened from `Autoskill/Autoskill Maker`.
After you click the `Done` button, autoskill command is copied to clipboard.

~~Order change doesn't work right now.~~
~~No way to set Enemy target right now.~~
~~Have to add Turn and Battle counters too.~~
UI is mostly a prototype.


reconman:
Attacking in the first turn without activating any skills produces `,#` instead of `0,#`

reconman:
How about "graying out" or darkening skill buttons if you've used them? Skills have at least 5 turns cooldown, and making them a different color will make it more easy to remember which ones have been used already.

When someone uses Order Change, the skills of the swapped-in servant could be turned normal again.

Also it could help people if they can put in Servant names instead of using "Servant 1", "Servant 2", "Servant 3", and so on. But make it optional so people can use the default Servant 1-6 if they can't bother to name their Servants. Servant 4-6 for Order Change.

But the biggest hurdle for all of my suggestions are Chen Gong and Arash who kill a Servant, so the dead Servant is replaced without Order Change.

MathewSachin:
> How about "graying out" or darkening skill buttons if you've used them? Skills have at least 5 turns cooldown, and making them a different color will make it more easy to remember which ones have been used already.

`Maid Alter / Tamamo (MLB KScope) / Tamamo (MLB Kscope)` would have to disagree :rofl:
I think we shouldn't show the cooldowns at all since even the script doesn't consider them.

>Also it could help people if they can put in Servant names instead of using "Servant 1", "Servant 2", "Servant 3", and so on. But make it optional so people can use the default Servant 1-6 if they can't bother to name their Servants. Servant 4-6 for Order Change.

Most people won't bother to enter the name of their servants, I think. It would instead make it more tiresome to use. They'd want to get this done quick and entering names of 3-6 servants to be used only once isn't golden.

> But the biggest hurdle for all of my suggestions are Chen Gong and Arash who kill a Servant, so the dead Servant is replaced without Order Change.
We would have to make a "kill" button and then use the Servant in the next sub member position to fill in the spot.

I'm against adding these kinds of things. People should at least try to remember which servants are in their team, which skills they used, etc. If we add that kill button, it's not necessary that people will click it at right time. If they can't remember, they can at least write it on paper or take a screenshot or something. These kinds of things can't be generalised.
You know, I had that idea of making skill command when a user plays a quests normally? I gave up on it after thinking that most users will (including me) impatiently click the screen during skill activations and stuff that it would be difficult to identify what the user pressed. So I instead went with this simplistic simulation.

MathewSachin:
Kotlin PR includes AutoSkill Maker marked as BETA

PR #104, opened 2020-05-07
Fix: Wrong region used for cropping Patterns

MathewSachin:
This caused problems with Support Selection like correct CE being selected with incorrect servant

### DIFF
--- a/CoreAutomata/Region.cs
+++ b/CoreAutomata/Region.cs
@@ -23,6 +23,19 @@ namespace CoreAutomata
             this.H = H;
         }
 
+        /// <summary>
+        /// Restrict another Region to be within the bounds of this Region.
+        /// </summary>
+        public Region Clip(Region Region)
+        {
+            var left = Region.X.Clip(X, R - 1);
+            var right = Region.R.Clip(X + 1, R);
+            var top = Region.Y.Clip(Y, B - 1);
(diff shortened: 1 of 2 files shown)


reconman:
The bug doesn't happen on Nox when debugging. I even tried setting the resolution to a non 16:9 scale, which changed nothing.

reconman:
Seems like it was a temporary bug that was fixed by restarting MediaProjection.

CoreAutomata/Region.cs
```diff
@@ -23,6 +23,19 @@ public Region(int X, int Y, int W, int H)
             this.H = H;
         }
 
+        /// <summary>
+        /// Restrict another Region to be within the bounds of this Region.
+        /// </summary>
+        public Region Clip(Region Region)
+        {
+            var left = Region.X.Clip(X, R - 1);
```
reconman, line 31: This will not work for Regions with X=0 and W=0, because `Clip` will return `-1`. X will be greater than the max value of `-1`.

This situation shouldn't happen, but we could add some Contract.Requires, for example in the setters of W and H. It would be like `Contract.Require(W > 0, "Region widths must be positive");`.

MathewSachin: I'm leaving it that way for now. And, from my experience, I'd rather avoid `Contract.Requires`, it needs to be activated or something to actually work.

PR #105, opened 2020-05-08
Trying Icons

MathewSachin:


reconman:
This is how the adaptive icons look like on my phone:
image

The icon is not perfect, but it's the best I could do without some complicated photoshop tricks.

MathewSachin:
Icons look great! Should we merge this or are any further changes required right now?

reconman:
It's fine to merge

PR #107, opened 2020-05-10
Update Xamarin, change target version to Android 10

reconman:
This may fix the Nox issues


MathewSachin:
Nope, sadly doesn't make a difference

PR #108, opened 2020-05-12
Kotlin

### LINKED ISSUE #106: White screen when opening the app.
First, I use Nox emulator. Last night, I can run the app normally with no issue. Then after I  update my Nox's android version to the newest ver., I can't open it anymore.

My friend too faced the same problem, and I am guessing its because the app is not supporting the newest android 7 ver. of Nox.

I have tried to create a new instance, but still facing the same problem. The same with re-install. Don't know if this is a bug or just some issue I and my friend faced. Really thank you if you fix this.

P.S. Waiting 10 min to see if the app will be opened after waiting that long, but nope.

MathewSachin:
Migrated the whole codebase to Kotlin.
In case you haven't heard of Kotlin, it is an alternative to JAVA for making Android apps and is now is the preferred language for Android development.

Also, fixes #106 which was related to Xamarin

### Other Changes
- Can record screen to MP4 file when the script is running.
- MLB star is matched with 0.77 similarity instead of 0.8. I noticed that it sometimes gets 0.78 similarity.
- AutoSkill maker is included
- Target Android 10 and added `android:requestLegacyExternalStorage="true"` to manifest
- Uses OpenCV 4
- User stopping the script is handled more gracefully now. Quick exit is possible from long waits too.
- Dark mode support for Android 9, 10
- AutoSkill delete button is on toolbar
- Clear Servant/CE preferences button is on the toolbar dropdown
- Extract default support images automatically on visiting support screen (if support image folder doesn't exist already)


app/build.gradle
```diff
@@ -0,0 +1,79 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
+
+android {
+    compileSdkVersion 28
```
reconman, line 6: Use Android 10
```suggestion
    compileSdkVersion 29
```

app/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+    buildToolsVersion "29.0.3"
+    ndkVersion "21.1.6352462"
+
+    defaultConfig {
+        applicationId "com.mathewsachin.fategrandautomata"
+        minSdkVersion 24
+        targetSdkVersion 28
```
reconman, line 13:
```suggestion
        targetSdkVersion 29
```

app/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+    maven { url "https://dl.bintray.com/blazei/maven" }
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation 'androidx.appcompat:appcompat:1.0.2'
+    implementation 'androidx.core:core-ktx:1.0.2'
+    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
+    implementation 'org.opencv:opencv-android:3.+'
```
reconman, line 69:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
    implementation 'androidx.appcompat:appcompat:1.1.0'
    implementation 'androidx.core:core-ktx:1.2.0'
    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
    implementation 'org.opencv:opencv-android:3.4.1'
```

app/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
+    testImplementation 'junit:junit:4.12'
+    androidTestImplementation 'androidx.test.ext:junit:1.1.1'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
+
+    def preference_version = "1.1.1"
```
reconman, line 75: Why are you using a variable for this?

app/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+    implementation "androidx.preference:preference-ktx:$preference_version"
+    implementation "androidx.coordinatorlayout:coordinatorlayout:1.1.0"
+    implementation 'com.google.android.material:material:1.0.0'
```
reconman, line 78:
```suggestion
    implementation 'com.google.android.material:material:1.1.0'
```

reconman:
I'll add more comments in the span of this week.

app/build.gradle
reconman: Notice, https://github.com/openpnp/opencv#api specifies that you have to do `OpenCV.loadShared()` before any native library can be used.

MathewSachin: Does this support Android though?
I knew about this library but the README doesn't mention Android support

app/build.gradle
MathewSachin: Google documentation suggested to add it like this. No other specific reason.

app/build.gradle
MathewSachin: If we target to Android 10, `Environment.getExternalStorageDirectory` is deprecated. It is the method used to find the path for `Fate-Grand-Automata` folder where support images are stored.
On Android 10, we can use `requestLegacyStorage` manifest option for now but won't work with Android 11.

app/build.gradle
MathewSachin: I decided to give this library a try anyway, but the app crashes on `OpenCV.loadShared()`.
I think it doesn't have Android support. The android specific OpenCV functions are also missing from the library.

bergerst: Hmm, but how will we ever get OpenCV updates? Do we really have to include the SDK in our build?

reconman: Hmm, but how will we ever get OpenCV updates? Do we really have to include the SDK in our build?

MathewSachin: we'll have to maintain a separate repo with OpenCV. I'll fork the library I'm currently using and update it to latest version of OpenCV.

MathewSachin: This one's quite up-to-date: https://github.com/iamareebjamal/opencv-android

app/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Battle.kt
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private fun autoChooseTarget() {
+        // from my experience, most boss stages are ordered like(Servant 1)(Servant 2)(Servant 3),
+        // where(Servant 3) is the most powerful one. see docs/ boss_stage.png
+        // that's why the table is iterated backwards.
+
+        var i = 2
+
+        for (target in Game.BattleTargetRegionArray) {
```
reconman, line 90:
```suggestion
        for (target in Game.BattleTargetRegionArray.reversed()) {
```

PR #110, opened 2020-05-14
WIP

### LINKED ISSUE #49: App is stuck in a loop if the inventory is full
The app will run in a loop if this screen pops up:

It should exit if this screen is detected.

MathewSachin:
I'll be working on this branch till the Kotlin PR (#108) gets merged.

- Better names for a few Activities, Fragments, etc.
- A message is shown on the AutoSkill list screen when there are no items.
- Fix: Withdraw region varies depending on whether you have Command Spells/Quartz.
- Make the Stop button RED when recording.
- Fix: App is stuck in a loop if the inventory is full (fixes #49).
- Automatically turn ON Ignore Notch calculations for Samsung users.
- Foreground service type: Media Projection.
- Card priority view is horizontal.
- Reset exitRequested before starting script (just in case)
- Ignore case when sorting Autoskill Configs by name

### Preferred Support
- Don't display .png after CE name in list.
- Renamed built-in CEs to more friendly names. This will only matter for new users or people who extract Support Images again.
- Ignore case when sorting preferred Servants/CEs.
- Handle blank values in preferred Support correctly
- Give priority to later servant ascensions

### AutoSkill Maker
- AutoSkill maker is shown as a button in Skill Command EditText dialog.
- Fixed a bug in Autoskill maker's `selectEnemyTarget`.
- Autoskill maker transitions are smoother.
- AutoSkill Maker Back button based on the current view.

### CN/TW
- Enabled use of `Game Server = CN/TW`
- Added the minimum necessary images to make the script function on CN/TW
- CN card type images were updated for the new card designs.
- Only uses images which are available based on the selected `Game Server` in result screen checks
- Lottery script displays error on CN/TW


app/src/main/java/com/mathewsachin/fategrandautomata/ui/MainActivity.kt
```diff
@@ -126,7 +126,11 @@ class MainActivity : AppCompatActivity(), PreferenceFragmentCompat.OnPreferenceS
                 // This initiates a prompt dialog for the user to confirm screen projection.
                 startActivityForResult(instance.mediaProjectionManager.createScreenCaptureIntent(), requestMediaProjection)
             }
-            else instance.start()
+            else {
+                if (instance.start()) {
+                    instance.showForegroundNotification()
+                }
+            }
```
reconman, line 133:
```suggestion
            else if (instance.start()) {
                instance.showForegroundNotification()
            }
```

reconman:
In the Skill Maker, the NP selection doesn't display the order of the selected NPs. Should we introduce `1st` `2nd` and `3rd` when you select them?

Also a field where you can see the current `SkillCommand` String would be nice because it reminds you of what you already put in.

reconman:
Found a bug by playing around:
1. Do a fresh installation (you could back up app settings with Titanium backup)
2. Set up Accessibility and accept MediaProjection
3. Click on the play button

Then the message `java.lang.IllegalStateException: dir.listFiles() must not be null` appears

reconman:
Also opening the Support settings crashes the app now, I already had that in the `kt` branch.

reconman:
The 2 last problems are fixed.

I just discovered another interesting thing: if you do a video recording and you're currently in vertical mode (home screen) and then open FGO, it will look like this: https://i.imgur.com/4qJ9P5C.mp4

MathewSachin:
I intentionally did it like that. We can't change the recording size during MediaProjection, so on rotation it automatically fits. So, I always start MediaProjection in landscape size.

PR #115, opened 2020-05-22
Show dialogs using Accessibility Service

MathewSachin:
Since showing dialogs from a service is disabled by some manufacturers, I'm creating a view similar to a dialog as an Accessibility Overlay.

- Instead of notification, we now get dialogs to show Script exit messages. The script still vibrates on exit.
- Error messages have a copy button which will copy the error to clipboard.
- Auto Skill picker is shown on script start if you have Auto Skill enabled. The last one you used is selected by default. The selected Auto Skill settings has been removed from the UI.


MathewSachin:
If Support Image Maker is shown this way, we won't have to worry about it not showing on some devices.
The code can also be much simplified then, we can generate the images into the app's private cache directory like Android recommends and move only the images (renamed) which the user wants to the `Fate-Grand-Automata/support` folder.
Option for generating Friend images can also be added without making it too bothersome for the user. Right now, they have to delete what they don't want. We can change it to keep what you want, ignore the rest.

PR #120, opened 2020-05-26
GitHub Pages Website

MathewSachin:
We really need a website with a button to download the app. Some users are not familiar with GitHub and are having trouble downloading the app. So, I quickly put up a basic website.

docsify is good and all, but it's more suited for programmer tools, we're more of an end-user app.

Screenshot (116)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 Auto-battle app for Fate/Grand Order Android.  
 
+Download from our [Website](https://MathewSachin.github.io/Fate-Grand-Automata).
+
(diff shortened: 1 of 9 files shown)


MathewSachin:
Let's just put up a website now, we can tweak it later.

reconman:
I'm still at work, so I can't do a full review, but you can build the same kind of cover page in docsify: https://docsify.js.org/#/cover

I personally hate writing HTML myself, so I prefer docsify's markdown => HTML translation.

MathewSachin:
Yeah, I saw cover pages, but we need to do api call to GitHub Releases to get the latest download link.

BTW, you can use markdown files on GitHub pages too. No need for any other libraries.

PR #128, opened 2020-05-31
Fix usage of DisplayMetrics

### LINKED ISSUE #121: Script won't play, immediately throws error "after Resolution Change"

Using v0.12.0

Done, app had been running fine previously.

**Describe the bug**
Script immediately stops after pressing play button (on support selection screen or battle screen) and throws up a bunch of code.

**Screenshots**
Screenshot_20200526-212133_Nova Launcher
Screenshot_20200526-212112_Fate_GO

 - Device model: Samsung Galaxy S9
 - Screen size: 2960x1440
 - Android Version: Android 10
 - RAM: 4GB

MathewSachin:
DisplayMetrics are re-read on starting the script or dragging play button (fixes #121).

Also fixes the problem of play button not moving freely on Nox.

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
@@ -33,6 +33,29 @@ class ScriptRunnerUserInterface(val Service: ScriptRunnerService) {
 
     val windowManager = Service.getSystemService(Context.WINDOW_SERVICE) as WindowManager
 
+    private val metrics: DisplayMetrics get() {
+        val res = DisplayMetrics()
+
+        windowManager.defaultDisplay.getRealMetrics(res)
+
+        return res
+    }
+
+    val mediaProjectionMetrics: DisplayMetrics get() {
+        val res = metrics
+
+        // Retrieve images in Landscape
+        if (res.heightPixels > res.widthPixels) {
+            res.let {
+                val temp = it.widthPixels
+                it.widthPixels = it.heightPixels
+                it.heightPixels = temp
+            }
+        }
+
+        return res
+    }
+
     private val layoutParams = WindowManager.LayoutParams().apply {
         type = overlayType
         format = PixelFormat.TRANSLUCENT
@@ -56,7 +79,6 @@ class ScriptRunnerUserInterface(val Service: ScriptRunnerService) {
     }
 
     private var scriptCtrlBtn: ImageButton
-    val metrics = DisplayMetrics()
 
     init {
         val inflater = LayoutInflater.from(Service)
@@ -68,18 +90,8 @@ class ScriptRunnerUserInterface(val Service: ScriptRunnerService) {
             it.setOnTouchListener(::scriptCtrlBtnOnTouch)
         }
 
-        windowManager.defaultDisplay.getRealMetrics(metrics)
-
-        // Retrieve images in Landscape
-        if (metrics.heightPixels > metrics.widthPixels) {
-            metrics.let {
-                val temp = it.widthPixels
-                it.widthPixels = it.heightPixels
-                it.heightPixels = temp
-            }
-        }
-
-        layoutParams.y = metrics.widthPixels
+        val m = metrics
+        layoutParams.y = maxOf(m.widthPixels, m.heightPixels)
     }
 
     fun show() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
```diff
@@ -33,6 +33,29 @@ class ScriptRunnerUserInterface(val Service: ScriptRunnerService) {
...
+
+        return res
+    }
+
+    val mediaProjectionMetrics: DisplayMetrics get() {
+        val res = metrics
+
+        // Retrieve images in Landscape
+        if (res.heightPixels > res.widthPixels) {
+            res.let {
```
reconman, line 49: Why do you even use `.let`?

Also, I searched for swapping variables in Kotlin and found this:
```
res.widthPixels = res.heightPixels.also { res.heightPixels = res.widthPixels}
```

MathewSachin: let, also, run, apply can be used to manipulate an item without referring to it again and again by name. There are some differences like you refer the object using `it` or `this`

Swapping using also doesn't look good to me, so didn't use that. Maybe there's some other way.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
```diff
@@ -33,6 +33,29 @@ class ScriptRunnerUserInterface(val Service: ScriptRunnerService) {
...
 
+    private val metrics: DisplayMetrics get() {
+        val res = DisplayMetrics()
+
+        windowManager.defaultDisplay.getRealMetrics(res)
+
+        return res
+    }
+
+    val mediaProjectionMetrics: DisplayMetrics get() {
```
reconman, line 44: Can you add a KDoc explaining what this is used for? My assumption is that it's used to only take landscape screenshots and recordings.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
```diff
@@ -112,30 +124,26 @@ class ScriptRunnerUserInterface(val Service: ScriptRunnerService) {
...
-        val rotate = rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_180
+        val m = metrics
 
-        val x = (if (rotate) {
-            metrics.heightPixels
-        } else metrics.widthPixels) - layout.measuredWidth
-        val y = (if (rotate) {
-            metrics.widthPixels
-        } else metrics.heightPixels) - layout.measuredHeight
+        val x = m.widthPixels - layout.measuredWidth
```
reconman, line 129: Can you explain this logic? Is `layout` only the button? So if the screen is 1280 pixels wide and the button is 100 pixels wide, x == 1180?

Please add KDoc to this method.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
MathewSachin: Yeah, the name has `mediaProjection` in it and there's a comment inside saying about using only landscape. KDoc feels redundant, still added.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
MathewSachin: Yup, just as you thought. I don't think it was even complicated. The code before this PR was of course. I added KDoc, and also changed the name of `layout` to `scriptCtrlBtnLayout`.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerUserInterface.kt
reconman: In my opinion, `it` is pretty ambiguous, it could mean anything, and `let` is only useful when you don't want to assign something to a variable. In this case using `res` is the same amount of effort as using `it`.

MathewSachin: Umm, `it` isn't ambiguous. It is a convention in Kotlin. If there is only one parameter to a lambda, it's name by default is `it`.
Yeah, `run` should probably be used here in place of `let`.
The benefit of this approach is that `temp` doesn't leak out of that scope.
Another benefit I can think is if we rename `res` later to something else, maybe a longer more descriptive name, the amount of change required will be minimal.

reconman: `temp` is also only visible inside of the `if`. I still find using `res` better, but you can merge if you want to keep `let`.

MathewSachin: I want to keep `let` 😅

PR #147, opened 2020-06-27
Screen detection improvements

reconman:
- Reorder methods
- Check for quest reward regularly
- Show refills during continues


reconman:
The git diff is a bit chaotic because I reordered the methods.

The most important changes are showRefillsUsedMessage(), isInQuestRewardScreen() and questReward()

Issue #150, opened 2020-07-02
Lottery Script for TW server


the3dsandwich:
Hi,

I came across this project a couple weeks back and has been using it in Gilfest 2018 (currently running in TW). Super thanks for this app to exist!

Now I want to see if I can help contribute to making the lottery script work for TW server. I'm new to Kotlin or android development but I can try learning it. Also, as I'm on TW serve I can help extract any images if needed.

Thanks again for this great app!

MathewSachin:
That's really great @the3dsandwich that you're willing to contribute.

Getting the Lottery script to work should be quite simple, even if you're new to Kotlin.

1. First, enable the lottery script for TW
https://github.com/MathewSachin/Fate-Grand-Automata/blob/cc63a13b623ee6bc18c97a8d6d09b49cc7dab6ab/app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoLottery.kt#L40

2. Then add images corresponding to:
https://github.com/MathewSachin/Fate-Grand-Automata/blob/master/app/src/main/assets/En/lottery.png
https://github.com/MathewSachin/Fate-Grand-Automata/blob/master/app/src/main/assets/En/StopGifts.png

For cropping the images, take the screenshot of the game in 1280x720.
and crop the regions based on:
https://github.com/MathewSachin/Fate-Grand-Automata/blob/cc63a13b623ee6bc18c97a8d6d09b49cc7dab6ab/app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoLottery.kt#L15-L16
Image dimensions are half of the script dimension, so take half of all these values when cropping
e.g.
```
finishedLotteryBoxRegion = Left = 287, Top = 430, Width=35, Height=50
```
I'd recommend to crop it smaller than this size but still within those bounds, we don't want the image matching to miss at edges due to fraction rounding issues when scaling on different devices.

---

On a sidenote, does the TW server has the free quest repeat feature yet? NA got that recently with the engine update and JP has had it for some time.
If it is there, maybe we can add support for that too just by putting in an image.

the3dsandwich:
Thanks for the reply!

TW unfortunately hasn't got the fq repeat feature yet. We're still waiting. 

[...]

PR #151, opened 2020-07-03
Added images to enable lottery script in TW

the3dsandwich:
Hi all,

I've added the images needed for image matching, and changed `AutoLottery.kt` so it won't throw errors anymore. I don't know how to compile and check whether the images will work, though.

Also I've uploaded the original screenshots to issue #150 so if I did anything wrong maybe someone can help crop the images?

Thanks!


reconman:
You can check yourself if it works. See the `All checks have passed`? The first one is a build script for the APK.

1. Click on `Details`
2. Click on `Artifacts 1` on the right
3. Click on the blue `apk` link
4. Install the apk on your device

It will install a 2nd app parallel to the normal one. The configuration will be the default one, not the one from the normal app, also you will have to give permissions again.

If your current version does not work, you can try fixing it yourself. Just push the change to your branch, wait until the APK is built again and then repeat steps 1 to 4. This will update the 2nd app, therefore keeping the configuration.

the3dsandwich:
I found that the image cropped isn't the right position and it won't work correctly. Let me re-crop it and push again

reconman:
Your images look fine to me. Don't forget to change the server region to TW again.

reconman:
It could be that the lottery finished image region is too small. You can check by enabling the debug mode in the advanced options. It will show red borders around the search regions.

the3dsandwich:
I think it's because the text in TW ("1個") is different from NA ("1"), so I can't just crop the "1". I'll keep testing.

reconman:
Instead of editing the image, I would widen the search rectangle. (`finishedLotteryBoxRegion`)

The parameters are in this order: X, Y, width, height

the3dsandwich:
I see, but it's ok as this may affect parts I can't repair. The commit `test lottery image` (ce92552) works after changing the lottery image, and present box full alert works too

reconman:
I mean, I changed the lottery region at around February for the Da Vinci Lottery. We don't know if the JP image works or not, since there was no lottery since then.

That's why I would rather move the X to the left, make the region wider, and keep the `1` as TW image.

You won't break anything as long as the new region includes the area of the old region.

the3dsandwich:
Alright, then let me test how much I'll need to widen the search rectangle. Also I'm thinking of copying the `1` from NA instead of cropping it again as they seem to be the same (in terms of font and size).

the3dsandwich:
After testing, the previous commit works pretty well. 
btw, what was the reason to change back the lottery region (or was it ever changed back)?

reconman:
> btw, what was the reason to change back the lottery region (or was it ever changed back)?

The lottery script didn't work in NA, so I changed the region. This was around the time I got into the old Lua codebase, so I have no idea why the previous value was set the way it was.

app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoLottery.kt
```diff
@@ -37,8 +37,8 @@ class AutoLottery : EntryPoint() {
 
     override fun script(): Nothing {
         when (Preferences.GameServer) {
-            GameServerEnum.Cn, GameServerEnum.Tw -> {
-                throw ScriptExitException("Lottery script doesn't support CN and TW servers right now.")
+            GameServerEnum.Cn -> {
+                throw ScriptExitException("Lottery script doesn't support CN server right now.")
```
reconman, line 41:
```suggestion
                throw ScriptExitException("Lottery script doesn't support the CN server right now.")
```

the3dsandwich:
> The lottery script didn't work in NA, so I changed the region.

I see. So before we have a chance to test if there's one configuration that works for all servers, we just have to test and see which region works for the current lottery event. I suppose it make sense...

reconman:
The current official version of Fate/Grand Automata can be downloaded from the master build in https://github.com/MathewSachin/Fate-Grand-Automata/actions as soon as it's done.

PR #152, opened 2020-07-04
Fix offsets for the support image maker in the Friend screen

### LINKED ISSUE #149: Support Image Maker doesn't work on the friend list after engine update
**Describe the bug**
So, after the engine update, Support Image Maker doesn't work on the friend list.
DW added support for locking friends, which shifted everything on the friend list screen to the right.

Should be pretty simple to fix. Just need to adjust the regions in Support Image Maker.

**Screenshots**
<kbd></kbd>

reconman:
Fixes #149

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/SupportImageMaker.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/SupportImageMaker.kt
@@ -44,8 +44,9 @@ class SupportImageMaker(private var Callback: (() -> Unit)?) : EntryPoint() {
         val searchRegion = Region(2100, 0, 370, 1440)
 
         val regionAnchor = ImageLocator.SupportRegionTool
-        // At max two Servant+CE are completely on screen, so only those
-        val regionArray = searchRegion.findAll(regionAnchor, supportRegionToolSimilarity).take(2)
+        // At max two Servant+CE are completely on screen, so only use those
+        val regionArray = searchRegion.findAll(regionAnchor, supportRegionToolSimilarity)
+            .take(2).toList()
 
         val screenBounds = Region(0, 0, Game.ScriptSize.Width, Game.ScriptSize.Height)
 
@@ -64,7 +65,7 @@ class SupportImageMaker(private var Callback: (() -> Unit)?) : EntryPoint() {
             }
         }
 
-        if (regionArray.count() == 0) {
+        if (regionArray.isEmpty()) {
             throw ScriptExitException("No support images were found on the current screen. Are you on Support selection or Friend list screen?")
         }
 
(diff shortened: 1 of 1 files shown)


reconman:
I'm not sure whether to use `*2` for script coordinates or not, so it's done one way in one line, and then another way in another line.

app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/SupportImageMaker.kt
```diff
@@ -32,78 +30,41 @@ fun getFriendImgPath(Index: Int): File {
...
+        // the servant and CE images are further to the right in the friend screen
+        val supportBoundX = if (isInSupport) 53 * 2 else 88 * 2
+        var supportBound = Region(supportBoundX, 0, 143 * 2, 110 * 2)
+        val searchRegion = Region(2100, 0, 370, 1440)
 
-        val searchRegion = Region(2100, 0, 300, 1440)
-        val regionArray = searchRegion.findAll(regionAnchor, supportRegionToolSimilarity)
+        val regionAnchor = ImageLocator.SupportRegionTool
+        // At max two Servant+CE are completely on screen, so only those
+        val regionArray = searchRegion.findAll(regionAnchor, supportRegionToolSimilarity).take(2)
```
MathewSachin, line 48:
```suggestion
        val regionArray = searchRegion.findAll(regionAnchor, supportRegionToolSimilarity)
            .take(2)
            .toArray()
```

`findAll` returns a `Sequence` instead of an `Iterable`. So, if we don't convert it into an array or list, the `count()` call below won't reuse the same computation and do it again.

app/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/SupportImageMaker.kt
```diff
@@ -32,78 +30,41 @@ fun getFriendImgPath(Index: Int): File {
...
+                extractServantImage(it, i)
+                extractCeImage(it, i)
+                extractFriendNameImage(supportBound, isInSupport, i)
             }
-
-            ++i
         }
 
-        if (i == 0) {
+        if (regionArray.count() == 0) {
```
MathewSachin, line 67:
```suggestion
        if (regionArray.isEmpty()) {
```

after my review comment above, this way seems more Kotlin like.

Otherwise, nice refactoring 👍

MathewSachin:
> I'm not sure whether to use *2 for script coordinates or not, so it's done one way in one line, and then another way in another line

That depends on whether we're taking an image from screen or cropping an already taken image.
Ideally, both should've been done in the same way, but:
- For Servant and CE, we're cropping an image which has already been taken from screen.
- For Friend, we're taking image directly from screen, so script coordinates.

reconman:
> That depends on whether we're taking an image from screen or cropping an already taken image.

I think you're misunderstanding. Why does the old code specify the supportBound with *2 instead of just using the multiplied values to begin with, for example 106 instead of 53*2?

MathewSachin:
Oh, now I get what you were asking.

There is no solid reason really. I just wrote it that way when I was tweaking the values. Feel free to change it to use the final multiplied values.

Issue #35, opened 2020-04-21
Support class selector


dekar63:
Right now the app scrolls through the class you are seeing in the moment of the execution. 

I suggest to include an option on the settings to indicate which class is the support to go directly to the desired column to scroll.

MathewSachin:
I tried doing this similar to #158, but the loading time for support screen mess things up.
We first need a sure-shot way to determine that support list has loaded.

PR #158, opened 2020-07-11
Party selection

### LINKED ISSUE #145: Select the team at the start of the farm
First, I want to thank you so much for your work on this project. It really made me appreciate the game a lot more.
Now for the request: I often switch teams to perform other quests (mainly story). So when I go back to the farming and launch fate grand automata, It happen quite often that the team used is not the correct one. Therefor a piece of code that check / switch the team used at the start of a battle would be cool. I know that I can resolv this issue just by thinking about it. But I think this feature would be a good quality of life addition to the project.

MathewSachin:
An option for party to use added to AutoSkill

fixes #145 

From: https://github.com/29988122/Fate-Grand-Order_Lua/pull/401


MathewSachin:
I'll be working next on #35 after this is merged.
Support class selection would be pretty similar to implement.

reconman:
There's one optimization potential: If you do quest repeats without the Continue button, you can skip the party selection on the 2nd quest run.

So we could introduce a 2nd boolean variable where we remember if we ever selected the party. This means that the app will switch parties if you start it in the middle of battles or in the result screen, but at least it won't select the party **every** time.

PR #162, opened 2020-07-12
Added logging

MathewSachin:
Some examples of the kinds of logs we get in Android Studio's Run dialog and logcat:

```
D/DroidCvPattern: match (En/retry.png on 350x150): 0.36141282320022583
D/<init>: click Location(X=1100, Y=675) x55
D/<init>: click Location(X=825, Y=560) x1
D/<init>: swipe Location(X=18, Y=595), Location(X=3, Y=330)
D/<init>: swipe Location(X=18, Y=595), Location(X=3, Y=330)
D/DroidCvPattern: match (ce/Chaldea Luchtime.png on 189x546): Match(Region=Region(X=16, Y=161, Width=158, Height=45), score=0.918035089969635)
D/DroidCvPattern: match (En/limitbroken.png on 8x45): 0.7202911972999573
D/DroidCvPattern: match (ce/Chaldea Luchtime.png on 189x546): Match(Region=Region(X=16, Y=361, Width=158, Height=45), score=0.9177476763725281)
```

These can be really helpful for checking the MLB issues.

I'll do logging in other areas after I can finalize the DI (#159) PR.


app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/AccessibilityGestures.kt
```diff
@@ -18,6 +19,8 @@ class AccessibilityGestures(private var AccessibilityService: AccessibilityServi
         swipePath.moveTo(Start.X.toFloat(), Start.Y.toFloat())
         swipePath.lineTo(End.X.toFloat(), End.Y.toFloat())
 
+        Log.d(AccessibilityGestures::class.simpleName, "swipe $Start, $End")
```
reconman, line 22:
```suggestion
        Log.d(this::class.simpleName, "swipe $Start, $End")
```

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/AccessibilityGestures.kt
```diff
@@ -38,6 +41,8 @@ class AccessibilityGestures(private var AccessibilityService: AccessibilityServi
             clickDuration.toLongMilliseconds()
         )
 
+        Log.d(AccessibilityGestures::class.simpleName, "click $Location x$Times")
```
reconman, line 44:
```suggestion
        Log.d(this::class.simpleName, "click $Location x$Times")
```

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -38,6 +50,10 @@ class DroidCvPattern(private var Mat: Mat? = Mat(), private val OwnsMat: Boolean
         Mat = null
     }
 
+    private fun logd(msg: String) {
+        Log.d(DroidCvPattern::class.simpleName, msg)
```
reconman, line 54:
```suggestion
        Log.d(this::class.simpleName, msg)
```

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -97,7 +118,10 @@ class DroidCvPattern(private var Mat: Mat? = Mat(), private val OwnsMat: Boolean
                         Template.height
                     )
 
-                    yield(Match(region, score))
+                    val match = Match(region, score)
+
+                    logd("match ($Template on ${this@DroidCvPattern}): $match")
```
reconman, line 123: What does this `this@DroidCvPattern` do?

app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
```diff
@@ -41,6 +42,8 @@ class RootScreenshotService(private val SuperUser: SuperUser) : IScreenshotServi
                         AutomataApi.PlatformImpl.toast("Unexpected raw image format: $format")
                     }
 
+                    Log.d(RootScreenshotService::class.simpleName, "${w}x${h} format=$format")
```
reconman, line 45:
```suggestion
                    Log.d(this::class.simpleName, "${w}x${h} format=$format")
```

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
MathewSachin: We're inside a lambda passed to the `sequence` function here. So, `this` here would refer to the `Sequence<Match>` object.
So, we have to qualify it with `this@DroidCvPattern` that we want the `DroidCvPattern` instance.

Due to this behaviour, I'm a bit sceptical about using `this::class.simpleName` in place of `[ORIGINAL_CLASS_NAME]::class.simpleName`.

MathewSachin:
> `D/DroidCvPattern: match (ce/Chaldea Luchtime.png on 189x546): Match(Region=Region(X=16, Y=161, Width=158, Height=45), score=0.918035089969635)`

We're matching with `Chaldea Lunchtime` as Template on a `189x594` image. I added a `tag` parameter to `IPattern` and filepath is passed as a tag and is used in the `toString` call. If there is no `tag`, image size is used. `189x594` is the size of target image.
The `Match` object printed after is the result. So, the size there is size of `Chaldea Lunchtime.png`.

> And why does `D/DroidCvPattern: match (En/limitbroken.png on 8x45): 0.7202911972999573` have just the number instead of score=

Because, in `exists`, we're not creating `Match` object, but in `findAll` we need to return them.

That `Match(Region=Region(X=16, Y=161, Width=158, Height=45), score=0.918035089969635)` is the `Match` object `toString()` value.

reconman:
I would like to make the logger messages more intuitive, for example:

`Matched ce/Chaldea Luchtime.png at X=100, Y=200 with a score of 0.95`
`Matched En/limitbroken.png with a score of 0.9`

The values themselves can be obtained using `${object.property}`.

Also according to the documentation of Log, Log.debug() will not be called in non-debug builds.

MathewSachin:
Hmm, these messages are much better.
I didn't notice it before, the coordinates in log message is relative to the base region. So, the region is within the image cropped from screen. To get the original region, we'll have to log from the `Region.findAll`, `Region.exists` methods instead of `DroidCvPattern`. I'll make this change too.

I knew, I added logging for our testing use only. Well, I won't expect people to run Android Studio or logcat to report logs.

MathewSachin:
Or, to keep things simple, we can drop the region from the logs. Debug mode shows the regions better anyway.
We're mostly interested in matching score.

PR #171, opened 2020-07-14
Custom sliders for delays/durations/similarity

### LINKED ISSUE #27: Introduce a wait multiplier for slow devices
The original Ankulua script has problems on some devices or emulators because it doesn't wait long enough.

Users should be able to influence the wait time, otherwise they have to build their own versions of the APK.

For this purpose, something like a wait multiplier should be introduced, which is used in every `AutomataApi.Wait()` call. The actual waiting time will then be `waitTime * waitMultiplier`.

 A value of < 1 should also be allowed for JP and for future NA versions with the new Unity engine.

### LINKED ISSUE #124: [JP] The First Support is always ignored
Already tested
 I'm not a samsung user

**Describe the bug**
The script/bot/app ignores the first support everytime

**Screenshots**
https://streamable.com/5zl3q0

**Device Info:**
 - Device model: Xiaomi Redmi note 7
 - Screen size: 1080 x 2340
 - Android Version: Android 9
 - RAM: 3GB

### LINKED ISSUE #136: Too much scrolling in support selection, resulting in missed supports

**Describe the bug**
The support selection seems to scroll too much, and often misses the right support.

**Screenshots**
https://streamable.com/fj0h5i

Here I've set Waver and Chaldea Lunchtime as my preferred support/CE, but you can see that it scrolls too much each time and misses it.

 - Device model: [Samsung A70 (SM-A705MN)]
 - Screen size: [1718x1080 (I tweaked it when I used FGOlua)]
 - Android Version: [Android 10]
 - RAM: [not sure]

Great work by the way! I'm am a FGOlua user but I've been trying to port everything over since the UI is so much nicer.

### LINKED ISSUE #170: Personalized Delays and Durations
How about adding a settings menu for adjusting delays and durations of the presses so that the app can better match each user's game load speed and such?

MathewSachin:
Added a page named `Fine-tune` which can be opened from `More options`.

fixes #170 

#### Options available
- Min Similarity
- MLB Similarity (fixes #124). Also reduced default MLB similarity to 0.7 which should handle most cases. Otherwise you can customize.
- Wait after clicking (ms)
- Click duration (ms)
- Delay b/w repeated clicks (ms)
- Wait after swiping (ms)
- Swipe duration (ms)
- Support Swipe multiplier (fixes #136)
- Wait multiplier (fixes #27)


IAmLoaf21:
How about calling it simply "Sliders"?

MathewSachin:
I don't think "Sliders" would be a good name.
People will want to know what it does rather than the UI controls present there.

PR #173, opened 2020-07-15
Change MLB images

MathewSachin:
MLB images were getting lower similarities and not matching sometimes.

We have transparent images but the version of OpenCV with transparent matching support hasn't been released yet.
limitBroken

So, we cropped from that image and raised MLB matching similarity to 0.8.

From my tests on NA, for MLB CEs it gives similarity >=0.88, for non-MLB <=0.72.
So, 0.8 is a reasonable threshold.


MathewSachin:
Closing in favour of #171

PR #174, opened 2020-07-15
Fix: Friend lock can cause Support Servant matching to fail

### LINKED ISSUE #157: Friend lock can cause Support Servant matching to fail
With the Engine Update, Friends can be locked. That causes a lock icon to appear on the servant image in support selection screen.
This would cause matching accuracy to fall possibly causing matching to fail.
friend lock

Support Image Maker should be updated for servant images (crop a bit more from left)
AND
Support servant images here should be cropped from left:
https://github.com/MathewSachin/Fate-Grand-Automata/tree/master/app/src/main/assets/Support/servant

MathewSachin:
If you lock your friends, a lock icon shows on the left of servant image, which can cause matching to fail.

Instead of modifying in-built images and Support Image Maker, which would need everyone to regenerate their images,
we crop out the part which can potentially have the lock.

fixes #157

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
@@ -247,8 +247,9 @@ class Support {
     private fun findServants(): Sequence<Region> = sequence {
         for (preferredServant in preferredServantArray) {
             // Cached pattern. Don't dispose here.
-            val pattern = loadSupportImagePattern(preferredServant)
-                .cropFriendLock()
+            val pattern = cropFriendLock(
+                loadSupportImagePattern(preferredServant)
+            )
 
             for (servant in Game.SupportListRegion.findAll(pattern)) {
                 yield(servant.Region)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
```diff
@@ -248,13 +248,27 @@ class Support {
...
 
             for (servant in Game.SupportListRegion.findAll(pattern)) {
                 yield(servant.Region)
             }
         }
     }
 
+    /**
+     * If you lock your friends, a lock icon shows on the left of servant image,
+     * which can cause matching to fail.
```
reconman, line 261:
```suggestion
     * which can cause matching to fail.
     *
```

app/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
```diff
@@ -248,13 +248,27 @@ class Support {
...
+     * Instead of modifying in-built images and Support Image Maker,
+     * which would need everyone to regenerate their images,
+     * crop out the part which can potentially have the lock.
+     */
+    private fun IPattern.cropFriendLock(): IPattern {
```
reconman, line 266: I feel like this method is a bit dangerous to be declared on IPattern. You don't want to execute this code for any random image, just for servant images. And you also don't want to execute this multiple times in a row.

PR #184, opened 2020-07-18
Improve Auto-detect Game Server

MathewSachin:


app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
```diff
@@ -175,12 +179,21 @@ class ScriptRunnerService : AccessibilityService() {
 
     override fun onInterrupt() {}
 
-    private var foregroundAppName = ""
+    private var currentFgoServer: GameServerEnum? = null
 
+    /**
+     * This method is called on any [AccessibilityEvent].
```
MathewSachin, line 185: You may want to write here that it's not for every AccessibilityEvent but only for once we've subscibed in `@xml/script_runner_service`.
Right now we've only subscribed for `TYPE_WINDOW_STATE_CHANGED`.

reconman: Done

MathewSachin:
I recall I didn't confirm for other servers.

@sleeping-player @ScathachSkadi does it auto-detect KR server?
@the3dsandwich does it auto-detect TW server?

If it auto-detects all servers, we don't need it as a setting in the UI.

sleeping-player:
Detection improvement is good,
but there is multination FGO runners.

the3dsandwich:
It auto detects server? I always double check in the settings anyway, and sometimes people have multiple server clients on so I think the option still has a reason to exist

sleeping-player:
And can this be tested with the build applied to the latest version?

MathewSachin:
This has been present since a few versions. Can even be tested on the PlayStore version.
It detects the version you're playing automatically.

Like, if you have both NA and JP installed, if you run it after starting FGO NA, it will automatically set to NA.
If you run after launching FGO JP, it'll set automatically to JP.
So, it's not a problem if people have multiple clients installed.

NA, JP and CN are detected correctly right now.
So, I just wanted to confirm if TW and KR are detected correctly, then we don't need the user to manually set the server at all. Hence, we don't need that UI.

MathewSachin:
To test you can set it to some other server (like NA). Then launch FGO, start the script, stop it. Come back to FGA settings to see if the server setting changed automatically.

ScathachSkadi:
Testing now. I hope I am not late.

sleeping-player:
Works good.
I have another Google account that can download FGO JP, so I could test it.
Run JP->Run a FGA script. Worked as JP.
In FGA JP state, Run KR->Run several any apps->Switch to KR->Run a FGA script. Worked as KR.
Fine.

ScathachSkadi:
Working fine here as well. I set it to NA but it automatically changed to KR. I didn't know this feature existed. I am glad I got to know this today.

the3dsandwich:
Not sure if I messed something up, but it doesn't switch to TW for me. Uploading screen recording...

MathewSachin:
So, KR is working fine, that's great.

@the3dsandwich you have to start the script for it to automatically detect the server.
Like when you run a script like Battle, at that time it'll switch the server automatically.

the3dsandwich:
Well the short answer is that I ran a script before and after starting up FGO TW, and neither of them triggered detection.

Long answer: I'm still trying to get a screen recording.

MathewSachin:
@the3dsandwich I think the TW package name is wrong in the code.

https://github.com/MathewSachin/Fate-Grand-Automata/blob/66a86bdd842f472eaaeb75f62e22d764e9f1b731/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/enums/GameServerEnum.kt#L7

Is this the right one? https://play.google.com/store/apps/details?id=com.xiaomeng.fategrandorder

the3dsandwich:
Ah I see! There are 2 versions of FGO TW (the MyCard apk version and the Play Store version). The pPlay Store version should be **com.xiaomeng.fategrandorder** and I guess **com.komoe.fgomycard** is the mycard version.

MathewSachin:
Can you tell me a bit more about how the MyCard version is different?
Is it just only distributing outside of PlayStore?

the3dsandwich:
To be honest I'm not sure, but from what I've heard it's an external APK that lets you buy SQ in a different way (mycard). Last time I checked it's still fully supported by FGO TW. The difference should only be how you pay the devs

MathewSachin:
It seems CN also has `com.bilibili.fatego` and `com.bilibili.fatego.sharejoy`

MathewSachin:
@the3dsandwich check if this build detects: https://github.com/MathewSachin/Fate-Grand-Automata/actions/runs/262769236

If it does, then we can safely remove the UI for Game Server setting.

the3dsandwich:
I tested and it seems to work just fine!

PR #192, opened 2020-07-20
Fix an exception in party preference

fa-gray:
The ListPreference it comes from produces a string, not an int


MathewSachin:
I messed up again 😅 
Thank you for fixing!

PR #200, opened 2020-07-24
Fix: Story Skip

### LINKED ISSUE #148: Story skip does not work

**Describe the bug**
I got everything setup and it does work if I am going to farm one stage but if I set it up to go through the story stages it never presses skip and if I manually press skip it does not tap the screen when it says "Quest cleared Get reward"
**Screenshots**
If an error occurred, you can find the message by expanding the app's notification.
It is recommended that you atleast add one screenshot of FGO running on your device.

**Video**
If relevant, include a video of the script running on your device with `Debug mode` ON (recording functionality is available within the app in `More options`).

Bluestacks, rooted

MathewSachin:
fixes #148 

OpenCV release v4.4 which adds support for matching transparent images.
So, we can now reliably match Story Skip image which needs to be transparent.

I've added Story Skip check to the outer loop so it doesn't matter whenever it comes up it can get matched.

I was getting around 0.77 similarity when matching, and 0.44 when not matching, so I've set the similarity at 0.7.

After this PR is merged, I plan to take more things out of the `result` function. Specifically, I want to make the app respond faster to the free quest repeat dialog.

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
@@ -92,16 +92,16 @@ class DroidCvPattern(
             val result = DisposableMat()
 
             if (Template.width <= width && Template.height <= height) {
-                alpha.let { mask ->
-                    if (mask != null) {
-                        Imgproc.matchTemplate(
-                            Mat,
-                            Template.Mat,
-                            result.Mat,
-                            Imgproc.TM_CCOEFF_NORMED,
-                            mask
-                        )
-                    } else Imgproc.matchTemplate(
+                if (alpha != null) {
+                    Imgproc.matchTemplate(
+                        Mat,
+                        Template.Mat,
+                        result.Mat,
+                        Imgproc.TM_CCOEFF_NORMED,
+                        alpha
+                    )
+                } else {
+                    Imgproc.matchTemplate(
                         Mat,
                         Template.Mat,
                         result.Mat,
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
@@ -12,17 +12,34 @@ import java.io.InputStream
 import kotlin.math.roundToInt
 import org.opencv.core.Size as CvSize
 
-class DroidCvPattern(private var Mat: Mat? = Mat(), private val OwnsMat: Boolean = true) :
-    IPattern {
-    constructor() : this(Mat())
+class DroidCvPattern(
+    private var Mat: Mat? = Mat(),
+    private val OwnsMat: Boolean = true
+) : IPattern {
+    private data class MatWithAlpha(val mat: Mat, val alpha: Mat?)
+
+    var alpha: Mat? = null
 
     private companion object {
-        fun makeMat(Stream: InputStream): Mat {
+        fun makeMat(Stream: InputStream): MatWithAlpha {
             val byteArray = Stream.readBytes()
             DisposableMat(MatOfByte(*byteArray)).use {
-                return Imgcodecs.imdecode(it.Mat, Imgcodecs.IMREAD_GRAYSCALE)
+                val decoded = Imgcodecs.imdecode(it.Mat, Imgcodecs.IMREAD_UNCHANGED)
+
+                // If there are 4 channels (RGBA), last one is alpha
+                val alphaChannel = if (decoded.channels() == 4) {
+                    Mat().apply { Core.extractChannel(decoded, this, 3) }
+                } else null
+
+                Imgproc.cvtColor(decoded, decoded, Imgproc.COLOR_RGBA2GRAY)
+
+                return MatWithAlpha(decoded, alphaChannel)
             }
         }
+    }
+
+    private constructor(matWithAlpha: MatWithAlpha) : this(matWithAlpha.mat) {
+        alpha = matWithAlpha.alpha
     }
 
     constructor(Stream: InputStream, tag: String) : this(makeMat(Stream)) {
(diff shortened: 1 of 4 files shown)


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -188,17 +189,7 @@ open class AutoBattle(
...
-                0.5.seconds.wait()
-                game.menuStorySkipYesClick.click()
-            }
-        }
-
-        10.seconds.wait()
-
+        // TODO: Move to outer loop
         // Quest Completion reward. Exits the screen when it is presented.
         if (game.resultCeRewardRegion.exists(images.bond10Reward)) {
```
reconman, line 194: Hold on, why is Bond 10 handled twice? One time above, one time below?

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -75,9 +92,23 @@ class DroidCvPattern(private var Mat: Mat? = Mat(), private val OwnsMat: Boolean
             val result = DisposableMat()
 
             if (Template.width <= width && Template.height <= height) {
-                Imgproc.matchTemplate(Mat, Template.Mat, result.Mat, Imgproc.TM_CCOEFF_NORMED)
-            }
-            else logd("Skipped matching $Template: Region out of bounds")
+                alpha.let { mask ->
```
reconman, line 95: Doesn't this let only create an alias for the `alpha` variable?

Why not use alpha directly?

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -75,9 +92,23 @@ class DroidCvPattern(private var Mat: Mat? = Mat(), private val OwnsMat: Boolean
...
+                    if (mask != null) {
+                        Imgproc.matchTemplate(
+                            Mat,
+                            Template.Mat,
+                            result.Mat,
+                            Imgproc.TM_CCOEFF_NORMED,
+                            mask
+                        )
+                    } else Imgproc.matchTemplate(
```
reconman, line 104: To keep the if and else similar, I would add curly braces here too.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -12,19 +12,36 @@ import java.io.InputStream
...
+                } else null
+
+                Imgproc.cvtColor(decoded, decoded, Imgproc.COLOR_RGBA2GRAY)
+
+                return MatWithAlpha(decoded, alphaChannel)
             }
         }
     }
 
+    private constructor(matWithAlpha: MatWithAlpha) : this(matWithAlpha.mat) {
```
reconman, line 41: Is there any situation where we would have a constructor call with MatWithAlpha and ownsMat = false? ownsMat is part of the default constructor.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
MathewSachin: IIRC, this was for the CE rewards after battle

reconman: I see, so bond10Reward is used for both. But this code won't work anymore anyways because of timing issues.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
MathewSachin: Because, it is a global variable, once you've checked for null, you can't be sure in the next line that it won't be set null again.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
MathewSachin: `ownsMat` is only used for screenshots where alpha doesn't matter

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
reconman: You're not checking for null. Normally you would use `let` this way:
`nullable?.let {}`

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
MathewSachin: yeah, that is why I added that TODO there

MathewSachin: I don't think it would have worked reliably before

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
MathewSachin: I'm checking for `null` inside `let`. Maybe try compiling yourself to get the intuition.

MathewSachin: Ok. my fault. OpenCV doesn't care about null 🤦

MathewSachin: removed the `let`.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
MathewSachin: done

Issue #102, opened 2020-05-04
Support specifying minimum skill level for servants


MathewSachin:
e.g. This is useful for SkaSka systems (DSS).
Skadi's battery needs to be 10 for most setups, and other skills also need to be at high levels.
We should at least get this done before she arrives on NA.

I've got an idea for this.
Now that we have masked image matching implemented, if we can somehow get the 10 images of skill numbers with transparency, we should be able to detect skill levels without depending on OCR which can get pretty unreliable.

reconman:
Even if we get this pseudo-OCR to work, what level of freedom do we allow? Should the support be only 10/10/10 or do we restrict it only to 1 or 2 skills? For example, the 2nd Merlin skill isn't important for farming.

MathewSachin:
We can give 3 fields for minimum levels which will be 1/1/1 by default, simple.
But, like either all CEs are Prefer MLB or not at a time. The selected skill levels will apply to all preferred servants.

reconman:
Where are you planning to get the numbers from? Photoshopping in that small scale is really difficult.

MathewSachin:
We got another problem now. Template matching of transparent images isn't working as expected. Story Skip button currently matches to almost anything 😢

MathewSachin:
I've implemented the option to individually set for the 3 skills, whether you want them to be maxed or them.
Implemented specific level check other than Lv10 would be too much work and make the support selection slower.
And, skill level 10 is what matters mostly like in DSS. So, closing this issue now. Expect the feature to be available in next release.

PR #204, opened 2020-07-25
Option to require that skills be Maxed (Lv10)

MathewSachin:
Solves part of #102 

Added a skill level page which can be opened from AutoSkill configs.
There you can specify if you want a skill to be maxed or not.

### Other UI changes
- MLB option is shown only if some CE(s) is/are selected
- `FriendsOnly` is forced for Friend mode and hence hidden in that case.

~~### Other (ideally should go to master before this PR)~~
~~Cropped, resized and copied images keep the same tag so that understanding the logs is easier.
Added logs for the score at which matching stopped in `findAll`.
This helped to find and fix a case with `floodFill` where the whole `result` `Mat` got filled to zero by using `FLOODFILL_FIXED_RANGE` flag (identified the problem by seeing that matching stopped at a score of zero).~~


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
```diff
@@ -393,4 +409,36 @@ class Support(fgAutomataApi: IFGAutomataApi) : IFGAutomataApi by fgAutomataApi {
         val mlbSimilarity = prefs.support.mlbSimilarity
         return limitBreakRegion.exists(limitBreakPattern, Similarity = mlbSimilarity)
     }
+
+    private fun skillLevels(bounds: Region, skillLevels: List<Boolean>): Boolean {
+        val y = bounds.Y + 325
+        val x = bounds.X + 1600
+
+        val skillLoc = listOf(
+            Location(x + 2 + 25, y),
```
reconman, line 418: Why areou you adding 2 and then 25 instead of doing `x = bounds.X + 1627`?

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
```diff
@@ -393,4 +409,36 @@ class Support(fgAutomataApi: IFGAutomataApi) : IFGAutomataApi by fgAutomataApi {
...
+
+        val skillString = result.withIndex().joinToString("/") {
+            val maxReq = skillLevels[it.index]
+            when {
+                !maxReq -> "x"
+                it.value -> "10"
+                else -> "f"
+            }
+        }
+        println(skillString)
```
reconman, line 440: Shouldn't this be logd?

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
MathewSachin: Because I was testing while adjusting the positions. 🤣. I'll change that before merging.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
MathewSachin: we don't have logd here. We're in an non-Android project 😢

MathewSachin: I didn't want to add an interface and pass it right now.

reconman: I thought about it back then, maybe we can get the same kind of logs with https://github.com/tony19/logback-android, which relies on the Slf4j API. Slf4j can also be used in Java modules.

MathewSachin: slf4j has an official library for Android: http://www.slf4j.org/android/
We don't need to use logback then.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
MathewSachin: Done

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
MathewSachin: last commit on slf4j-android was 7 yrs ago 😢

reconman: But still, nothing speaks against using it, if it works.

MathewSachin: I'm a bit confused. Maybe they're using some different repo now.
Last release was in Dec'19
https://mvnrepository.com/artifact/org.slf4j/slf4j-android

reconman: Beware of this: https://stackoverflow.com/a/35012588/5971497

MathewSachin: yeah, I'd seen that issue. It's for Verbose. We're logging at debug level, so it should be fine.

reconman: In the linked Slf4j JIRA ticket, there's a reference to this library, which is also getting updates and actually logs debug messages: https://gitlab.com/mvysny/slf4j-handroid

reconman: > It's for Verbose

Nope, I've read that everything below INFO is not logged.

MathewSachin: yeah, you're right. I tried running it and didn't log. I'll try `slf4j-handroid` then.

MathewSachin: done

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Support.kt
```diff
@@ -393,4 +412,38 @@ class Support(fgAutomataApi: IFGAutomataApi) : IFGAutomataApi by fgAutomataApi {
...
+            if (!skillLevels[it.index])
+                true
+            else {
+                val skillRegion = Region(it.value, Size(35, 45))
+                skillRegion.exists(images.skillTen, Similarity = 0.68)
+            }
+        }
+
+        logger.debug {
+            // Skill command as string
```
reconman, line 436:
```suggestion
            // Detected skill levels as string for debugging
```

PR #208, opened 2020-07-27
Use NA images as fallback for other servers

MathewSachin:
I missed adding the Skill level 10 check image for servers other than NA.
That's when I thought, why not have other servers fallback to NA for images.

Removed images from other servers which were just straight copies of NA images.
Tested on NA and JP (Nox)


app/src/main/java/com/mathewsachin/fategrandautomata/util/ImageLoader.kt
```diff
@@ -57,14 +60,29 @@ class ImageLoader @Inject constructor(
...
     }
 
+    /**
+     * When image is not available for the current server, use the image from NA server.
+     */
+    private fun loadPatternWithFallback(path: String): IPattern {
+        if (currentGameServer != GameServerEnum.En) {
+            return try {
+                createPattern(currentGameServer, path)
+            } catch (e: FileNotFoundException) {
```
reconman, line 78: I don't really like the catch because it's bad for the performance, here's a blog article about C#: https://www.c-sharpcorner.com/blogs/impact-of-trycatch-on-performance

The alternative involves going through all assets in the gameserver folder:
```
if (context.assets.list("$currentGameServer").contains("$path")) {
  //load from specific gameserver folder
}
else {
  //load from NA
}
```

app/src/main/java/com/mathewsachin/fategrandautomata/util/ImageLoader.kt
```diff
@@ -31,13 +32,15 @@ class ImageLoader @Inject constructor(
         return null
     }
 
-    private fun createPattern(FilePath: String): IPattern {
+    private fun createPattern(gameServer: GameServerEnum, FileName: String): IPattern {
+        val filePath = "$gameServer/${FileName}"
```
reconman, line 36: Is the `${FileName}` necessary or would `$FileName` suffice? You're not calling any method or reading properties here.

app/src/main/java/com/mathewsachin/fategrandautomata/util/ImageLoader.kt
MathewSachin: I don't think it really matters, since we do it only once per asset and then cache it.
But, yeah, I agree with the suggestion, but you'll have to open a separate PR. I already merged this 😅.

Issue #137, opened 2020-06-16
Selecting a specific servant's face card


kelpie2:
Before I get started, I'd just like to say that the app looks fantastic, and seems to work well for the most part. It is certainly a lot more user-friendly than FGOlua, which I've used in the past.

I'd like to request an option to prioritise selecting a specific servant's face card during card selection. This would be especially useful for something like double Skadi with Dantes, where to finish off a beefier wave 3 boss you might want to rely on Dantes' boosted quick card crits. Since Dantes hits most targets neutrally, the weak/netural/resist prioritization wouldn't prioritize his card over Skadi's cards in most cases.

I'd imagine this would require uploading an image of the servant's face card like you to for support selection. I'm hope I'm not asking for something that is too difficult to implement.

MathewSachin:
Implementation aside, for these kinds of features, we need to decide the UI we'll use for configuration. It should be optional since everyone won't want this level of customization.
And if we add support for detecting servant face-cards, there are other features some users want like brave chains and for cases like farming bones with face-cards, we'd like to avoid brave chains for fast clears.
We need to come up with a UI which allows setting this up without too much of a hassle.
This is kind of a disadvantage of having a UI over code I'd say. We have to keep it simple enough for most users.
I'm open to suggestions for this. Do post here if you have an idea in mind.

metapone:
If the UI would be too cluttered adding this feature, how about doing something like [ArchiSteamFarm](https://github.com/JustArchiNET/ArchiSteamFarm): make a dedicate UI (on app/web?) to generate config file for users to download and copy into their phones?

tchofy:
Hmm, maybe something like the Support Image Maker, or use the command card sprites from fgo wiki (fandom), although not all servants have them, unfortunately, and the app would need transparent image detection support, which I'm not sure if the library they use got updated to include that already. Then I guess every Auto-Skill script would need a Card Priority setting that overrides the one in the homescreen.

I did a quick mockup in photoshop that addresses brave chain issues, 
![mockup](https://user-images.githubusercontent.com/64620317/85352737-08679680-b4dd-11ea-8f1a-3374aab078c6.jpg)

If so happens that you drew 5 cards for the same servant, the script will just fallback to the default card priority on the homescreen. Don't think it's a good solution, and it involves a lot of coding, but I guess it's a start.

MathewSachin:
@tchofy We now have Card priority per battle:
<kbd>
[...]

PR #212, opened 2020-07-29
Option to prefer or avoid brave chains (Experimental)

MathewSachin:
I'm thinking of keeping the `Experimental` label since these features don't work in all cases.

A `Brave Chains` setting added in `More options` with the following possible values:
- Don't care
- After NP (Doesn't care about the Cards before NP)
   Attempt towards resolving #137 
  Tracks the first NP used from AutoSkill and gives priority to the same servant's face cards.
  Uses as many cards of the same servant is available while respecting Card Priority settings after which it falls back to the normal Card Priority handling.
  Doesn't work with NP Spam/Danger
- Avoid
   Avoids brave chains if possible. Useful if you're face-card farming (e.g. bones) or want to finish up remaining health of 2 enemies after NPs.

### Rearranging cards (Configurable)
When clicking 3 cards, move the card with 2nd highest priority to last position to amplify its effect
 (A,B,C -> A,C,B) when A > B > C
Do the same when clicking 2 cards unless they're used before NPs.

Skip if:
- NP spamming because we don't know how many NPs might've been used.
- in Avoid brave chain mode because it tries to get consecutive cards to be not of the same servant.


reconman:
Why would you avoid brave chains?

reconman:
> Detects if NPs are loaded using Edge detection: This part needs tweaking. Danger mode doesn't seem to work even outside this PR?

When I last tested this 2-3 months ago, it was still working.

MathewSachin:
> When I last tested this 2-3 months ago, it was still working.

Yeah, used to work earlier. I only tested in the summer nodes where it didn't work for me.

MathewSachin:
Oh, does Danger mode only work if Auto-targeting is ON? Seems so from the code.

MathewSachin:
I'm merging this PR now since it is getting difficult to test multiple features with the restriction on AP.
Shouldn't be much of a problem since both of the features in this PR are marked as Experimental and are opt-in.

reconman:
Sorry, I was quite busy the past few days.

MathewSachin:
Yeah, I understand. I wanted to test this with #231. That seems to be working pretty good on NA. JP doesn't have an event right now, so wasn't able to get an image of `Unable to find support matching criteria` screen yet.

PR #219, opened 2020-08-01
Remove unnecessary party switch

MathewSachin:


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -321,16 +322,30 @@ open class AutoBattle(
         val party = prefs.selectedAutoSkillConfig.party
 
         if (!partySelected && party in game.partySelectionArray.indices) {
-            // Start Quest Button becomes unresponsive if the same party is clicked.
-            // So we switch to one party and then to the user-specified one.
-            val tempParty = if (party == 0) 1 else 0
-            game.partySelectionArray[tempParty].click()
+            val currentParty = game.selectedPartyRegion
+                .findAll(images.selectedParty)
```
reconman, line 336:
```suggestion
                .findAll(images.selectedParty)
                .take(1)
```

More than 1 match should never happen.

reconman: Alternatively we could implement a findFirst() method, which just calls `return findAll().take(1)` in case we need this more often.

reconman:
I did some comment changes just now

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
MathewSachin: `take(1)` doesn't make a difference here since I'm calling `firstOrNull`.
It is a `Sequence` and therefore, is lazily evaluated.

MathewSachin: Also, `take(1)` would still return a `Sequence` until a terminal operator like `firstOrNull` is used.

reconman: Didn't know the last part, I thought it immediately maps every value instantly.

PR #228, opened 2020-08-04
Made friend pt summon script only do friend summons

shineburst:


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoFriendGacha.kt
```diff
@@ -1,9 +1,20 @@
...
 import com.mathewsachin.fategrandautomata.scripts.IFGAutomataApi
+import com.mathewsachin.fategrandautomata.scripts.prefs.IPreferences
 import com.mathewsachin.libautomata.*
 import kotlin.time.seconds
 
+fun IFGAutomataApi.isInFriendPtsSummon(){
+    val startRightSwipeLoc = Location(300, 500)
+    val endRightSwipeLoc = Location(2000, 500)
+
+    while (!game.friendPtSummonCheck.exists(images.friend, Similarity = 0.3)) {
```
MathewSachin, line 12: Something's definitely wrong if you have to use such a low Similarity of 0.3.
The default similarity we use is 0.8!
I think the friend icon is of different size on the summon screen. I'll check with a new image.

MathewSachin: Yup, the image was of different size. I've pushed a new image.
And its working great on NA! I'll test on JP in a few minutes.

MathewSachin: works on JP too!

PR #231, opened 2020-08-05
Waiting till support screen loads

### LINKED ISSUE #129: Wrong friend support picked if update list takes too long due to bad internet.
When the update list takes too long loading, or gives connection error, the script will try to update list again instead of retrying. This is ok if the error is shown, but since the update list action is time based (i think); if the script tries to update the list just before the previously called update finishes, it will end up touching a random support servant when it tries to hit the 'ok' button. This happens more than you'd think on my shitty internet.

This is a great project and I hope I was able to help you improve it a little.

### LINKED ISSUE #164: Repeat Quest will not wait long enough for the support list to load
In 70 % of the cases, the app will update the support list after clicking on the Continue button for repeating the quest. Seems like we need to increase the wait time between selecting `Continue` and checking the support list.

### LINKED ISSUE #229: After the first or second battle the script keeps refreshing the friend list even though the selection mode is set to "first"
Sorry if my english is bad.

I have the latest release

> Samsung user? See https://mathewsachin.github.io/Fate-Grand-Automata/SAMSUNG.html

I am using NOX

**Describe the bug**

After the first or second battle the script keeps refreshing the friend list even though the selection mode is set to "first". 
The script works fine the first time but it bugs after that. 
After it refreshes the list for like 2 to 3 minutes it finally selects a support. 

**Video**

https://www.youtube.com/watch?v=dprjYbfxDXk

 - Device model: NOX
 - Screen size: 720x1280
 - Android Version: Android 7
 - RAM: 4GB

MathewSachin:
Seems to be working fine on NA.

fixes #229
fixes #164 
fixes #129


MathewSachin:
For First support selection mode, this code doesn't handle the case where a friend hasn't put up a support servant in a slot. Well, if you're sorting by Level/HP/Atk this shouldn't happen, but can happen if sorting by Login.

support not set

MathewSachin:
Need images of this screen for JP, CN, TW:

support not found

PR #236, opened 2020-08-06
Option to take Screenshot of drops (Experimental)

### LINKED ISSUE #188: Auto Reward Screen Screenshot Toggle
It would be handy for logging purposes, such as contributing to Atlas Academy, if FGA can automatically screencap the drop screen when toggled.

MathewSachin:
fixes #188 

When this option is ON, the Result screen is handled by a separate branch of code.


MathewSachin:
@homu-gamepress can you try the APK from this page: https://github.com/MathewSachin/Fate-Grand-Automata/actions/runs/197847815

There's a new option in `More options` you have to turn ON.

homu-gamepress:
I got it to take a couple drop screenshots, but the current NA event mechanics is making it difficult to make a mass of runs. Will report back once I have a bit more to test with.

homu-gamepress:
I tested a dozen runs on Ice Bridge and it seemed to screencap correctly:

https://imgur.com/a/HoaLTOB

Is everything supposed to be greyscale?

Will try to test whether it scrolls down correctly on large drop counts, once I can get a auto-able setup working for the event.

MathewSachin:
> Is everything supposed to be greyscale?

yup, grayscale is used everywhere in the script because it's faster than having to deal with 3 colour channels.

reconman:
I think that the guys over at Atlas Academy can't deal with grayscale at the moment.

homu-gamepress:
The second drop screenshot is a bit high, maybe move the scroll down click down a few pixels?

homu-gamepress:
Tried out the new version and it seemed to fixed the issue!

homu-gamepress:
Click version is much better!

image


Previous scroll version:

image

MathewSachin:
@homu-gamepress Do you think this is good enough?
I just feel that it is too slow, not that I can think of any way to make it faster.

homu-gamepress:
It's fast enough. 

Bigger issue to testing whether greyscale screenshots causes any misreading for the drop image parser. Any way to take color screenshots?

MathewSachin:
> Bigger issue to testing whether greyscale screenshots causes any misreading for the drop image parser. Any way to take color screenshots?

I can take a color screenshot, if it is full-screen including those blue-borders and notch.
Cropping it to just FGO for color screenshots too would require some parts of the code to be rewritten.

homu-gamepress:
Yeah, full screen works just fine for AA.

MathewSachin:
Haven't checked screenshot using root, but for the MediaProjection one, I've checked that color screenshots are being taken.

homu-gamepress:
Yep, it's in color! Awesome job!

image

MathewSachin:
I've confirmed that this works fine on Nox with Screenshot using root and JP server.
I guess if I match the scrollbar, we can avoid the unnecessary scroll and duplicate screenshot if there aren't that many drops.

PR #245, opened 2020-08-08
Fix: 10x summon (continue) misclick

potchy:
The old location for `continueSummonClick` didn't even register on my phone, so I moved it slightly up:
Screenshot_20200807-123004_Fate_GO

[Before](https://www.youtube.com/watch?v=E5b1LFgE8_s) and [after](https://www.youtube.com/watch?v=lkoki9sQddg) video comparison. I'm not sure if JP's layout is different. D:


MathewSachin:
You're right, it's even below the point where it clicks continuously. I'm wondering why it even works on my phone. 🤣

MathewSachin:
JP layout is pretty much the same, so there won't be any problems.

PR #249, opened 2020-08-12
Separate instance of ExitManager per script run

MathewSachin:
Every script run having a different ExitManager would avoid race-conditions.
ExitManager internally uses a CoroutineScope to manage cancellation. This would allow introducing parallelism in the code later.
And, due to the CoroutineScope, instance exit from waits is possible.

Also, updated a few libraries and refactored the Dagger code.


libautomata/src/main/java/com/mathewsachin/libautomata/ExitManager.kt
```diff
@@ -1,38 +1,41 @@
...
  */
-@Singleton
-class ExitManager @Inject constructor() {
-    @Volatile
-    private var exitRequested = false
+class ExitManager {
+    private val scope = CoroutineScope(Dispatchers.IO)
+
+    fun wait(duration: Duration) {
+        checkExitRequested()
```
reconman, line 13: Is checkExitRequested() now only done once before waiting?

MathewSachin: No, it's still being used before things like Region.exists, Region.findAll etc.

MathewSachin: I'm just not launching the coroutine if the scope is already cancelled.

reconman: I meant that we don't manually check ourselves every second that someone stopped the script. But I just now read that scope.cancel() deals with terminating threads.

MathewSachin: yup, I misread your comment.
It also exits immediately unlike the once per second check.

PR #309, opened 2020-08-27
Better handling of Bond/Reward CEs

MathewSachin:
If the script somehow clicked on a Bond CE when clicking rapidly, it can get stuck there.
This fix adds another check for the CE details screen.


reconman (approved):
Can't test it using my gallery app because of the notch, but the code looks fine.

MathewSachin:
There way my gallery app works, I was able to test even with a notch

PR #348, opened 2020-09-02
Play notification sound when script stops

### LINKED ISSUE #263: Notification sound when the app gets stuck or exits
Currently as it stands, the app gets stuck in some screens like Bond 10 (#262), when new nodes appear but aren't automatically selected by the game (#211), and those are conditions that are really hard to test and handle, so my proposal is the following:

If either the app automatically stops because of an error, or if it hasn't done any clicking for a period of 1~2 mintues (depending on the device, maybe more), it should notify the player using audio, so that if the player is near their phone they'll know something went wrong, and for those who go to sleep and leave their phone farming during the night (Gilfest is close) they won't wake up with the surprise their X servant reached max bond 10 minutes after they slept and lost the whole night's worth of farming.

MathewSachin:
~~Haven't tested on Android 7 yet. Charging my old phone right now so that it at least boots.~~

Tested on Android 10 which should be same as for 8 and 9.

A notification is sent by the app when a script stops, which expires after 10 seconds. The notification won't be intrusive and will just play a sound.
This will also happen when the user turns OFF the screen to stop the script or when the screen turns OFF by itself after the system display timeout setting, which can happen if a script is stuck at some screen.

On Android 8+, you can configure notification audio, vibration and a few other parameters from the Android's notification settings for the app itself, so it saves a lot of work for me. The `Messages` channel is used for these notifications. If someone is annoyed by these notifications, they could even turn them OFF.

fixes #263 
@tchofy can you test the build from this PR?

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerNotification.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerNotification.kt
@@ -1,9 +1,6 @@
 package com.mathewsachin.fategrandautomata.accessibility
 
-import android.app.NotificationChannel
-import android.app.NotificationManager
-import android.app.PendingIntent
-import android.app.Service
+import android.app.*
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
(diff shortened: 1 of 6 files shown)


tchofy:
Running it on Android 10, at first it wasn't working, I was getting a vibration, a notification, but no sound, so I did some extra digging and found out it was my rom's (Oxygen OS on a Poco F1) game mode blocking it's sound. Now it works as expected when it stops.

PR #350, opened 2020-09-03
Korean Translation

ScathachSkadi:
Added new strings and translated some. Will translate rest after I wake up.

### DIFF
--- a/app/src/main/res/values-ko/localized.xml
+++ b/app/src/main/res/values-ko/localized.xml
@@ -5,33 +5,33 @@
     <string name="auto_skill_item_export">내보내기</string>
     <string name="auto_skill_item_export_failed">내보내기 실패</string>
     <string name="auto_skill_item_copy_name">%s (복사)</string>
-    <string name="auto_skill_item_delete_confirm_title">Confirm Deletion</string>
-    <string name="auto_skill_item_delete_confirm_ok">Delete</string>
-    <string name="auto_skill_item_delete_confirm_message">Are you sure you want to delete this configuration?</string>
-    <string name="auto_skill_item_clear_preferred">Clear</string>
-    <string name="auto_skill_item_open_maker">Maker</string>
-    <string name="auto_skill_support_any">Any</string>
+    <string name="auto_skill_item_delete_confirm_title">삭제 확인</string>
+    <string name="auto_skill_item_delete_confirm_ok">삭제</string>
+    <string name="auto_skill_item_delete_confirm_message">이 설정을 정말로 지우시겠습니까?</string>
+    <string name="auto_skill_item_clear_preferred">초기화</string>
+    <string name="auto_skill_item_open_maker">메이커</string>
+    <string name="auto_skill_support_any">아무거나</string>
 
-    <string name="support_imgs_extracted">Support Images Extracted Successfully</string>
-    <string name="support_img_not_found">Unable to load image: %1$s. Put images in %2$s folder</string>
+    <string name="support_imgs_extracted">서포트 이미지가 성공적으로 추출되었습니다</string>
+    <string name="support_img_not_found">이미지 불러오기 실패: %1$s. %2$s 폴더에 이미지를 넣으시오.</string>
 
     <string name="auto_skill_list_no_items">오토 스킬 아이템 없음</string>
     <string name="auto_skill_list_import">불러오기</string>
     <string name="auto_skill_list_export_all">모두 내보내기</string>
     <string name="auto_skill_list_delete">삭제</string>
-    <string name="auto_skill_list_selected_count">%d selected</string>
-    <string name="auto_skill_list_export_failed">Failed to export %d item(s)</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-ko/localized.xml
```diff
@@ -142,8 +142,8 @@
     <string name="p_boost_item">부스트 아이템</string>
     <string name="p_brave_chains">브레이브 체인 (실험적)</string>
     <string name="p_rearrange_cards">카드 재배열</string>
-    <string name="p_auto_skill_cmd_text">AutoSkill Cmd TextBox</string>
-    <string name="p_auto_skill_cmd_text_summary">Show text-box for editing AutoSkill commands</string>
+    <string name="p_auto_skill_cmd_text">오토 스킬 Cmd 텍스트 박스</string>
```
reconman, line 145: Shouldn't `Cmd` also be translated as `커맨드를`?

app/src/main/res/values-ko/localized.xml
```diff
@@ -159,23 +159,23 @@
 
     <string name="p_fine_tune">미세 조정</string>
     <string name="p_fine_tune_default">기본값: %d</string>
-    <string name="p_fine_tune_support">Support</string>
+    <string name="p_fine_tune_support">서포트</string>
     <string name="p_fine_tune_support_swipes_per_update">Swipes per Update</string>
```
reconman, line 163: Not translated

app/src/main/res/values-ko/localized.xml
ScathachSkadi: Oh that was the short form of command. Right. Will change that and also what does b/w mean? Black and white?

reconman: No, in this context it's between. I've never heard of the abbreviation myself, though.

ScathachSkadi: Me neither. Thanks for that.

ScathachSkadi:
https://del.dog/imeappocop.txt

Crashes when you try to edit the command of auto skill

reconman:
@MathewSachin Seems like this error is caused by https://github.com/MathewSachin/Fate-Grand-Automata/blob/master/app/src/main/java/com/mathewsachin/fategrandautomata/ui/auto_skill_maker/AutoSkillMakerViewModel.kt#L373

See https://stackoverflow.com/a/59633268/5971497

I tried renaming it to `clearCommand` in Android Studio, but now I'm getting compiler errors, it can't resolve `AutoSkillMakerActivityArgs`.

app/src/main/res/values-ko/localized.xml
MathewSachin: Reworded `AutoSkill Cmd TextBox` to `Skill Command TextBox`.
If I put `between` in place of `b/w` the text becomes too long. @reconman Should I reword `Delay between repeated clicks` to `Repeat click delay`?

MathewSachin:
@reconman I renamed the clear function and tried a release build. It seems like the problem is something else.

This is happening only in release builds, so, Proguard is removing something again.

> it can't resolve AutoSkillMakerActivityArgs.

It'll resolve once you build. It's compile-time generated code.

MathewSachin:
@reconman You were right. My bad. It was a permission problem on my phone.

MathewSachin:
@ScathachSkadi @reconman I've confirmed that the crash is fixed in the latest GitHub Actions build for this PR

ScathachSkadi:
Awesome.

ScathachSkadi:
@MathewSachin is there something that I need to retranslate or it is the same as before?

MathewSachin:
@ScathachSkadi Nothing new to translate. Just check that all screens render properly, in case I missed something.

app/src/main/res/values-ko/localized.xml
reconman: Maybe `Time between clicks`

MathewSachin: We need to make it clear that it is between repeated clicks somehow.

reconman: @ScathachSkadi We decided on Multi-Click Interval for the English version.

ScathachSkadi: Well I think the translation itself would change in that case

ScathachSkadi:
https://ibb.co/xKZkSQK
I don't think Gold x0 should be seen here. The rest of the app has no problems so far said the people who I gave the forked apk

MathewSachin:
You'll have to live with that `Gold x0`. It just prints the `enum` value as it is.

ScathachSkadi:
Oh okay. People are just starting to test it out so I will let people use the app a bit more to see if there's something wrong with the translation. I am going to try the app irl myself to kill some Hassans.

MathewSachin:
I guess I'll merge this now. You can open another PR if you want to make more changes.

PR #356, opened 2020-09-04
WIP: Back up in the foreground

reconman:
This PR is just so I can test the behavior during Gilfest.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -9,6 +9,7 @@
     <application
         android:name=".util.AutomataApplication"
         android:allowBackup="true"
+        android:backupInForeground="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"


MathewSachin:
From [documentation](https://developer.android.com/guide/topics/manifest/application-element):

> **android:backupInForeground**
Indicates that Auto Backup operations may be performed on this app even if the app is in a foreground-equivalent state. **The system shuts down an app during auto backup operation, so use this attribute with caution.** Setting this flag to true can impact app behavior while the app is active.
The default value is false, which means that the OS will avoid backing up the app while it is running in the foreground (such as a music app that is actively playing music via a service in the startForeground() state).

Doesn't seem like a good idea since the app will be killed for backup.

reconman:
When I tried this build, the play button never disappeared. But I didn't test it for long.

MathewSachin:
But, do we really need this now? 🤔 
AutoSkill configs are the only things that matter much and we have Import/Export for that.

I'm more worried because the documentation clearly says that this setting allows Android to kill a foreground app for backup.

I think you might not have encountered a problem because AutoBackup doesn't happen all the time.
They also state that the conditions for backup would usually be met every night, but, we have a lot of users who leave the app running overnight. The app stopping in between runs won't be nice.

MathewSachin:
In the Google Drive app, open Backups from the menu, pick your device, then App Data, then you can see when each app was last backed up.
Both PlayStore and Debug builds are getting backed up regularly (almost every night) for me even without this manifest option.
And, the Accessibility Service also restarts automatically after the backup on my phone. Should be the case with most phones unless like Xiaomi which prevents app startup unless explicitly allowed by the user.

MathewSachin:
I've checked that its being backed up consistently right now. We don't want to risk the app being killed when it is running, so closing this PR.

reconman:
Forgot to comment, I have 0 backups for FGA in Google Drive right now.

Have to try again with a foreground backup build.

MathewSachin:
One thing to note, in the Google Drive app, backups for apps not installed from PlayStore show on the bottom and with their complete package name, unlike PlayStore apps.

reconman:
I already looked there

MathewSachin:
One thing I want to confirm, do you leave FGA as a foreground service overnight?

reconman:
Yes, I do most of the time.

MathewSachin:
Just turn it OFF at night, it'll get backed up lol.
Allowing a foreground app to be killed is a bit dangerous.

reconman:
I'll try out this branch version and tell you how it behaves. I'm always lazy about stopping the FGA service, that's why I requested the `hide play button in vertical` feature. 😃

MathewSachin:
Now that we know that the exact reason the Accessibility problems happen, allowing the app to be killed while in the foreground would only worsen the problem. Closing again ...

PR #359, opened 2020-09-06
Friend Point Summon image for TW server and increased click when continuing summons

the3dsandwich:
Noticed that fp_continue.png is missing for TW server so I added it, but this needs some testing so I've created this PR so I can test on my phone. Not sure if only adding the image is enough for it to work


MathewSachin:
Hey, have you set up Android Studio? That'll allow you to test directly on your phone which would be much easier than having to wait for CI build to finish and download the zip of the APK.
The scripts also print some logs which can be seen in Android Studio which help in debugging, like what similarity a particular image is getting.

the3dsandwich:
@MathewSachin Yes I did, but I haven't really looked into how to get the built apk on my phone so now I'm testing both with the CI and on the emulator.

Also, I noticed there needs to be one more click after clicking summon (to confirm summon), so I added that too.

MathewSachin:
> Yes I did, but I haven't really looked into how to get the built apk on my phone

You'll have to turn ON Developer options and USB debugging on your phone. Then it'll show in Android Studio. You can then click the Play button in Android Studio and the app will build and launch on your phone.

> Also, I noticed there needs to be one more click after clicking summon (to confirm summon), so I added that too.

I think TW might be on the old Friend Summon format. NA got updated to the JP format where you can continue straight from the result screen.
So, make that extra click only if GameServer == TW.

the3dsandwich:
@MathewSachin Thanks for the tips for setting up Android Studio! It's working correctly on my pc now. I've also changed the part to only give an extra click when game server is TW, but personally I can verify whether it will create bugs for other servers, so I'll need help on checking (in theory it should work)

MathewSachin (approved):

MathewSachin:
It should keep working for other servers. No need to worry 👍.

PR #364, opened 2020-09-07
Support Friend Summon for KR server

MathewSachin:
Friend summon for KR also seems to use the old format. So, doing it similar to #359.

@sleeping-player @ScathachSkadi can you check this build: https://github.com/MathewSachin/Fate-Grand-Automata/actions/runs/242964064?


sleeping-player:
Oh wait

sleeping-player:
I'm sorry for leaving you do it as I was late for what I should have done..

As I already did daily free fp summon today, so I couldn't test it. It seems to work fine except for this exception.

MathewSachin:
Daily summon doesn't really matter since it's only done once.
Summoning continuously is more important for using up FP.

ScathachSkadi:
Checking now

ScathachSkadi:
Runs fine. I also used my 10 fp gacha before testing it.

ScathachSkadi:
10 free fp gacha doesn't work. Just tried. But I don't think that matters too much.

PR #367, opened 2020-09-08
Wording Fixes for TW Localization

the3dsandwich:
Some of the text needs reviewing. hopefully this time around we can catch all the errors and weird word choices I've made.


app/src/main/res/values-b+zh+TW/localized.xml
```diff
@@ -12,7 +12,7 @@
     <string name="auto_skill_item_open_maker">產生器</string>
     <string name="auto_skill_support_any">任何</string>
 
-    <string name="support_imgs_extracted">成功擷取支援從者圖片</string>
+    <string name="support_imgs_extracted">成功截取支援從者圖片</string>
```
reconman, line 15: Is 截取支 really correct? Online translation tools translate it to to English as `interception`. Are words like 提取 or 解压 better? What is done here is that the images are copied from the APK to the file system.

the3dsandwich: I see. Then how about using just "copy" `複製` for it?
It's different from the support image maker, right?

the3dsandwich: I'll also change `<string name="support_menu_extract_default_support_images">截取預設支援從者圖片</string>` as well

MathewSachin: it's different from support image maker.
Only pictures are extracted.
The default images contained in the app for Merlin, Waver, Skadi, etc.

the3dsandwich (commented):
`support_imgs_extracted` changed to (rough translation) `image copy success` (圖片複製成功)
`support_menu_extract_default_support_images` changed to (rough translation) `copy default support image` (複製預設支援圖片)

As app is copying images I think using "copy" may be better to reflect the result

the3dsandwich:
It would be nice if we can have another TW user take a look at the translation. To be honest I'll still be using the English interface even after we fix all the text problem.

By the way, when is Gilfest for NA?

reconman:
> It would be nice if we can have another TW user take a look at the translation.

You're the only TW user of FGA, I think. You would have to ask around in taiwanese FGO forums.

> By the way, when is Gilfest for NA?

Around September 20th.

MathewSachin:
> Around September 20th.

Does that mean we get a dead week?

reconman:
> > Around September 20th.
> 
> Does that mean we get a dead week?

Yes, probably. But I just noticed that Fate/Zero started 5 days early, so Gilfest may also start early. Or they give us a few days of break.

the3dsandwich:
Well then. If there happens to be any FGO TW user around and found the translation text unacceptable, then I suppose we can fix it then. Somehow TW forums are quite strict about assisting tools, so that's sad.

TW will have Christmas from Sep 11th. I guess I'll be doing my best to record videos (when AP won't be a problem) and seeing if anything need tweaking.

the3dsandwich:
On this note, just realized I have the CEs for Gilfest still lying around. I can help add some of them in if you don't mind

MathewSachin:
NA doesn't need to make images of event CEs anymore. We already got the JP feature of filtering by event CE

reconman:
> TW will have Christmas from Sep 11th.

Are all seasonal events not at the time they should be? 

> NA doesn't need to make images of event CEs anymore. We already got the JP feature of filtering by event CE

Which doesn't fully work in this event, for some reason.

MathewSachin:
> Which doesn't fully work in this event, for some reason.

Fate/Zero? Maybe you have the Bonus servant setting ON.

MathewSachin:
And other CEs were in the general pool. So, they're not really event CEs

the3dsandwich:
> Are all seasonal events not at the time they should be?

I don't know about other servers, but we on TW do Christmas in summer and have dead weeks that lasts 3 months.

> NA doesn't need to make images of event CEs anymore. We already got the JP feature of filtering by event CE

Oh, right. Almost totally forgot about this feature. So I guess I don't need to help add that (?) Cuz I noticed the CE for summer event is still there so thought it's still needed

the3dsandwich:
(Gilfest2020) CKT
(Gilfest2020) Return Match
(Gilfest2020) Shining Skyscrapers

I'll put them here (the ones I have) just in case

reconman:
@MathewSachin can we remove https://github.com/MathewSachin/Fate-Grand-Automata/blob/master/app/src/main/assets/Support/ce/Painting%20Summer.png?

MathewSachin:
I added Painting Summer because it is one of those 50% starting gauge general purpose CEs.
We don't bundle event CEs in the app otherwise.

People can use Support image maker is they really need those images.

the3dsandwich:
Ah I see. No problem with CEs then

PR #368, opened 2020-09-09
Use a different icon for debug versions

reconman:
Icons were generated, they just need to be configured to be used for debug builds. I haven't looked into the details on how to do that.


MathewSachin:
Icons added in #369

PR #374, opened 2020-09-11
When both Servants and CEs are preferred, look for CEs in a more confined location

### LINKED ISSUE #372: Support selection sometimes ignores CE setting.
I am. 0.20.0
> Samsung user? See https://mathewsachin.github.io/Fate-Grand-Automata/SAMSUNG.html
I mean, the emu *reports* as a Samsung device, but it's Bluestacks, so no.

**Describe the bug**
Something I've been running into every now and then is that when I set both a preferred support and a preferred CE the script sometimes straight up ignores the CE requirement, tapping the first servant that meets the requirements, regardless of CE. I've tried finding the cause of this behavior and, while I can't be sure, I believe the answer may lie in the difference between standard CE images and user-created ones. 

I tested it with the preset Waver and Lunchtime images, as well as custom-made Angra Mainyu and Before Awakening (apparently I misremembered the name) ones.
The script has no issues when I set the requirements to a Waver with any CE/Lunchtime, Angra with any CE/Lunchtime, or any servant with HRP. but when I ask it to find a specific servant (custom or preset) with BA, it'll click the first eligible servant in the list, regardless of CE. It does still properly take other limitations into account, like friends only or max skill settings.

**Screenshots**
N.A.

**Video**
I don't think screenshots or videos are particularly relevant here, as the behaviour isn't different from what is expected, at least visually.

 - Device model: [Bluestacks version 4.220.0.1109]
 - Screen size: [1600x900]
 - Android Version: [7.1.2]
 - RAM: [4 GB]

p.s. the built-in Lunchtime CE image seems to be misspelled as Luchtime. Not sure if that has ever been brought up but I figured I'd mention it anyway.

MathewSachin:
fixes #372

### DIFF
--- a/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Game.kt
+++ b/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/Game.kt
@@ -66,6 +66,7 @@ class Game @Inject constructor(val prefs: IPreferences) {
 
         val supportRegionToolSearchRegion = Region(2100, 0, 370, 1440)
         val supportDefaultBounds = Region(76, 0, 2356, 428)
+        val supportDefaultCeBounds = Region(76, 270, 378, 150)
         val supportExtraRegion = Region(1200, 200, 130, 130)
         val supportNotFoundRegion = Region(468, 708, 100, 90)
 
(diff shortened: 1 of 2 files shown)


reconman (approved):
Looks good, I looked at the red borders and they seem correct.

PR #407, opened 2020-09-15
Gift Box script

### LINKED ISSUE #397: Auto Retrieve Silver EXP from the Gift Box
I saw that you said that you won't implement the Auto Gift Box feature from FGOLua due to missing Text Recognition but I was wondering if it's possible to make it automatically retrieve Silver EXP cards of any quantity since that is image recognition and not text. Great app either way!

MathewSachin:
This will select all (max 99) Silver embers and [Gold embers based on a threshold set in more options (name subject to change)]
The font for numbers seems different depending on the server in question.

JP and NA working with the same images.

fixes #397


MathewSachin:
@ScathachSkadi I've added KR images. A build you can test should be ready in a few minutes.

ScathachSkadi:
As of now, I only have gold exp cards but I'll try it

ScathachSkadi:
It works well.

MathewSachin:
@the3dsandwich I've added TW images. Give it a try.

the3dsandwich:
On it.

the3dsandwich:
@MathewSachin the scroll amount is too much. It will skip items without checking some of the items with each scroll now.
I'm uploading the debug video. Should finish uploading in a few minutes.

the3dsandwich:
I tweaked a little bit and found that swiping from `Location(700, 1200)` to `Location(700, 650)` (changed to Y:650) gives a pretty nice result

MathewSachin:
Hmm, right the scrolling wasn't very smooth with those values.
Let me try your values on NA too.

MathewSachin:
I got so busy replying to all questions and issues that I forgot to reply.
@the3dsandwich your swipe value works much better!

PR #443, opened 2020-09-20
Quest end images for TW

### LINKED ISSUE #442: Quest Completed Screen not detecting
Before opening an issue:

> Your issue will get closed/locked by the bot in case proper info isn't provided
If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

**Describe the bug**
In Battle, if the current screen is item drop (the one you want to screenshot), if you con't decide to screenshot drops and somehow was stuck on the screen, FGA won't detect it nor click to skip it.

**Screenshots**
Screenshot_1600561986

**Video**
[recording](https://youtu.be/RP_w0QdlS2Q)

 - Device model: [Pixel 2 API 30(Android Studio Emulator]

the3dsandwich:
Fixes #442


MathewSachin:
Is this ready to merge?
I was planning on a release to fix problems on LG phones, so maybe I could include this too.

the3dsandwich:
Still trying to get master_exp working. Somehow the cropped images just won't trigger...

MathewSachin:
See if this works:

image

the3dsandwich:
Ended up cropping it again and it seems to work. Checking it again to see if the matching is reliable enough (it should be).

the3dsandwich:
Yup, it works. Wasn't really sure what made the difference (tried cropping from Windows image viewer, Photoshop, but letting the image sit for a couple hours and reloading a couple times seemed to be more effective than changing cropping app)

MathewSachin:
I usually just crop in MS Paint 🤣 

What I do is:
1. Remove blue/black borders around FGO.
2. Resize to 1280x720
3. Check regions from the script. Image dimensions are half of the script dimensions. For example, if the script uses `(10, 20, 30, 40)`, we'll crop out `(5, 10, 15, 40)` from the image. These values are `(Left, Top, Right, Bottom)`.
4. I also usually try to crop out a bit smaller region that what is specified in code, so we don't have any cases of the image going out of region bounds on very large or very small phones.

MathewSachin:
I think the image size was the problem, the earlier image was too big compared to the current one.

Issue #457, opened 2020-09-21
Apply Support Swipe Multiplier for all swipes


reconman:
The Gift Box script doesn't scroll enough on my phone. Normally I have the support swipe multiplier set to 141 and that value alone fixes the support scroll swiping.

The Gift Box script doesn't use that multiplier. If I lower the swipe duration, it will only swipe 50 % of the time. If I increase the swipe duration, it doesn't make a difference. So modifying the swipe time does not help me.

Issue #430, opened 2020-09-19
Screenshot Drops - If Drops are more than 42, it will miss 1 line of the drop


nFeyz:
The apps only take the screenshot top of the drop page and end of the drop page, but if the drops are more than 42 drops (JP 90+ node), it will miss 1 line of the materials. Just want to let you guys know of this (will undercount if we submit the drop from the screenshot).
![18-9-2020-09-31-37 0](https://user-images.githubusercontent.com/53361797/93657561-5d786180-fa66-11ea-9954-d8b2f64b74b4.png)
![18-9-2020-09-31-37 1](https://user-images.githubusercontent.com/53361797/93657568-65380600-fa66-11ea-8b00-bd9d9edb2cee.png)

MathewSachin:
Yeah, I know this is the case.
When the scrollbar isn't shown, we take 1 screenshot. If the scrollbar is there, we scroll to bottom and take one more.
For, the 3 screenshot case, detecting that there are 3 screenshots worth of drops is difficult

nFeyz:
What if the apps scrolls the drops by dragging the materials? (like how you scroll the support) and detect if its at the end of the page or no by matching the end of the bar? 
![image](https://user-images.githubusercontent.com/53361797/93659023-7d625200-fa73-11ea-9197-b7ef943b9ac0.png)
![image](https://user-images.githubusercontent.com/53361797/93659041-b8648580-fa73-11ea-859d-9b68ce7e80e2.png)

MathewSachin:
Yeah, that was what I was thinking. But, scrolling in this app isn't very precise, so there's no avoiding getting duplicate screenshots. I mean some of the drops will be shared between the screenshots since it's not possible to scroll exactly one page everytime.

PR #459, opened 2020-09-23
Add Support Swipe Multiplier to Gift Box

the3dsandwich:
Solves #457 

For now I think copying and modifying `scrollList()` from `Support.kt` is enough, though you may be interested in making swiping in a separate module so other scripts can refer to it?

Let me prepare a working screen recording before this can be merged.

### DIFF
--- a/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoGiftBox.kt
+++ b/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoGiftBox.kt
@@ -23,7 +23,6 @@ class AutoGiftBox @Inject constructor(
     override fun script(): Nothing {
         while (clickCount < maxClickCount) {
             checkGifts()
-
             scrollList()
         }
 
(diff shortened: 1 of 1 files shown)


the3dsandwich:
[Screen Recording](https://youtu.be/o6aJbJHgxy0)

MathewSachin:
#430 is one more case where we would potentially need to swipe, so I was thinking of renaming `supportSwipeMultiplier` to `swipeMultiplier` and putting the linear interpolation code into the `swipe` function.

But, for that, the `swipeMultiplier` should go into [IPlatformPrefs](https://github.com/MathewSachin/Fate-Grand-Automata/blob/master/libautomata/src/main/java/com/mathewsachin/libautomata/IPlatformPrefs.kt)

Also, the interpolation is only done along Y-axis right now. We need to do it along X-axis too for it to be generalized.

the3dsandwich:
Yeah I think that is a better way to implement swipes. But if you need a quick and dirty way to get gift box working before the generalized solution then I guess here's this.

reconman:
I fully support what Mathew proposed. I will use your version for now, but I'd rather not put something into the master branch temporarily.

MathewSachin:
No problem. I'll take this up.
@the3dsandwich would this be the right translation for `Swipe multiplier`: `滑動乘`?
In Korean, the words were separate, so I could make sense, but not sure about TW.

the3dsandwich:
Let me think about the translation. The wording of 滑動乘 doesn't sound natural enough.  I'm thinking along the lines of 滑動微調 (slide tweak/fine-tuning) or 滑動加乘 (slide multiplication). Let me sit on both names for a day or so and see which one sticks

MathewSachin:
Closing in favour of #464 

Swipe multiplier would apply to all swipes

I rewrote how scrolling works on Android 8 and above, so we might not need to set Swipe multiplier at all for them.

PR #519, opened 2020-10-16
Hide PLAY button in vertical orientation

### LINKED ISSUE #379: Make toggle service button reflect the on/off state
Screenshot_20200912-183442
Maybe make this button reflect the on/off status

### LINKED ISSUE #518: Hide play button when in vertical mode
Just a new idea I came up with. When you stop the script for some reason and want to look something up outside of FGO, then you have to constantly move the play button around so you can type and click properly or you have to stop the service.

What if the play button is only shown in landscape mode, which is automatically started when FGO is in the foreground?

MathewSachin:
fixes #518

PLAY button hides in portrait orientation and shows up in Landscape orientation.

#### Should we pause on switching to Portrait orientation?
If you switch out of FGO (into some other app in vertical orientation) while a script is running, it should ideally PAUSE. Right now it keeps running. The problem here is handling the Screen turning OFF. The scripts have to stop and notify the user that the screen turned OFF.
But, most phones will switch to Portrait orientation on the lock screen, which can lead to a race-condition. I'd rather avoid doing this.
Or, we could change the behaviour on the screen turn OFF to PAUSE the script, but also show the notification and play sound (#523).
Even after #523, I don't think this would work well.
Orientation change to portrait event is received before (at least on my phone) Screen turned off event. The screen turn off message would get lost this way, which is bad.

```
D/FGA: Detected FGO: En
V/FGA: PORTRAIT
V/FGA: SCREEN OFF
```

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
@@ -49,6 +49,8 @@ class ScriptRunnerService : AccessibilityService() {
             return (Instance?.start(mediaProjectionToken) == true).also { success ->
                 if (success) {
                     _serviceStarted.value = true
+
+                    Toast.makeText(Instance, "Open FGO to start scripts", Toast.LENGTH_SHORT).show()
                 }
             }
         }
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
@@ -26,7 +26,6 @@ import com.mathewsachin.fategrandautomata.scripts.prefs.IPreferences
 import com.mathewsachin.fategrandautomata.util.*
 import com.mathewsachin.libautomata.IPlatformImpl
 import com.mathewsachin.libautomata.IScreenshotService
-import com.mathewsachin.libautomata.ScriptAbortException
 import com.mathewsachin.libautomata.messageAndStackTrace
 import dagger.hilt.android.AndroidEntryPoint
 import timber.log.Timber
@@ -163,7 +162,7 @@ class ScriptRunnerService : AccessibilityService() {
     }
 
     private fun stop(): Boolean {
-        scriptManager.stopScript(ScriptAbortException.User())
+        scriptManager.stopScript()
 
         serviceState.let {
             if (it is ServiceState.Started) {
@@ -209,7 +208,7 @@ class ScriptRunnerService : AccessibilityService() {
 
             if (state is ServiceState.Started) {
                 when (scriptManager.scriptState) {
-                    is ScriptState.Started -> scriptManager.stopScript(ScriptAbortException.User())
+                    is ScriptState.Started -> scriptManager.stopScript()
                     is ScriptState.Stopped -> {
                         scriptManager.startScript(this, state.screenshotService, scriptComponentBuilder)
                     }
@@ -220,7 +219,7 @@ class ScriptRunnerService : AccessibilityService() {
 
     fun registerScriptPauseBtnListeners(scriptPauseBtn: ImageButton) =
         scriptPauseBtn.setOnClickListener {
-            scriptManager.togglePause()
+            scriptManager.pause(ScriptManager.PauseAction.Toggle)
         }
 
     override fun onServiceConnected() {
@@ -237,10 +236,11 @@ class ScriptRunnerService : AccessibilityService() {
         Instance = this
 
         screenOffReceiver.register(this) {
-            scriptManager.scriptState.let { state ->
-                // Don't stop if already paused
-                if (state is ScriptState.Started && !state.paused) {
-                    scriptManager.stopScript(ScriptAbortException.ScreenTurnedOff())
+            scriptManager.pause(ScriptManager.PauseAction.Pause).let { success ->
+                if (success) {
+                    val msg = getString(R.string.screen_turned_off)
+                    platformImpl.notify(msg)
+                    platformImpl.messageBox(msg, msg)
                 }
             }
         }
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
@@ -7,6 +7,7 @@ import android.app.PendingIntent
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Intent
+import android.content.res.Configuration
 import android.media.projection.MediaProjectionManager
 import android.os.SystemClock
 import android.view.accessibility.AccessibilityEvent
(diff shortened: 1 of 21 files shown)


reconman:
When you lock your screen while in FGO and then unlock it, the play button is gone until you switch to another app and back.

MathewSachin:
> When you lock your screen while in FGO and then unlock it, the play button is gone until you switch to another app and back.

On my phone PLAY button still displays, regardless of whether I unlock directly with fingerprint or by drawing the pattern in portrait mode.

reconman:
Seems like it was a one-time glitch. Worked every time when I tried again.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
```diff
@@ -254,8 +241,17 @@ class ScriptRunnerService : AccessibilityService() {
...
-                    ?.let { currentFgoServer = it }
+                val gameServer = GameServerEnum.fromPackageName(foregroundAppName)
+
+                if (gameServer != null) {
+                    prefs.gameServer = gameServer
+
+                    if (serviceState is ServiceState.Started) {
+                        userInterface.show()
+                    }
+                } else if (userInterface.metrics.let { it.widthPixels < it.heightPixels }) {
```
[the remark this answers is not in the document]
reconman:
```suggestion
                } else if (userInterface.metrics.let { it.widthPixels < it.heightPixels }) {
                    // Hides the overlay in portrait mode
```

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
```diff
@@ -49,6 +49,8 @@ class ScriptRunnerService : AccessibilityService() {
             return (Instance?.start(mediaProjectionToken) == true).also { success ->
                 if (success) {
                     _serviceStarted.value = true
+
+                    Toast.makeText(Instance, "Open FGO to start scripts", Toast.LENGTH_SHORT).show()
```
reconman, line 53: Shouldn't this text be localized?

reconman (changes requested):

MathewSachin:
I spammed screen lock/unlock and the PLAY button sometimes does not show like you said.

Correct case:
```
V/FGA: Detected FGO En
D/WindowManager: Add to mViews: com.mathewsachin.fategrandautomata.ui.HighlightView{2707390 V.ED..... ......I. 0,0-1080,2292}, this = android.view.WindowManagerGlobal@e7a7232,pkg= null
V/ViewRootImpl: The specified message queue synchronization  barrier token has not been posted or has already been removed
D/WindowManager: Add to mViews: android.widget.FrameLayout{2b39caf V.E...... ......ID 0,0-162,162}, this = android.view.WindowManagerGlobal@e7a7232,pkg= null
V/ViewRootImpl: The specified message queue synchronization  barrier token has not been posted or has already been removed
V/FGA: Horizontal: com.coloros.floatassistant
V/FGA: Horizontal: com.android.systemui
V/FGA: Horizontal: com.mathewsachin.fategrandautomata.debug
V/FGA: Horizontal: com.mathewsachin.fategrandautomata.debug
```

Faulty case:
```
V/FGA: Detected FGO En
V/FGA: Vertical: com.coloros.floatassistant
V/FGA: Vertical: com.android.systemui
V/FGA: Horizontal: com.coloros.floatassistant
```

System UI glitched a bit when that happened. I guess this can be solved if instead of checking vertical, we check for horizontal (after an FGO opened event has been received).

MathewSachin:
I changed the behaviour to: 'PLAY button hides in portrait orientation and shows up in Landscape orientation'.
And this seems to work fine because it's independent of FGO now.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
```diff
@@ -213,9 +226,19 @@ class ScriptRunnerService : AccessibilityService() {
     override fun onServiceConnected() {
         Timber.info { "Accessibility Service bound to system" }
 
+        // We only want events from FGO
+        serviceInfo = serviceInfo.apply {
```
reconman: So you're setting the package names when the accessibility service starts so you don't have to edit the script_runner_service.xml on package name changes, right?

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
```diff
@@ -230,28 +253,40 @@ class ScriptRunnerService : AccessibilityService() {
...
+     * of the FGO APKs.
      */
     override fun onAccessibilityEvent(event: AccessibilityEvent?) {
         when (event?.eventType) {
             AccessibilityEvent.TYPE_WINDOW_STATE_CHANGED -> {
                 val foregroundAppName = event.packageName?.toString()
                     ?: return
 
-                val gameServer = GameServerEnum.fromPackageName(foregroundAppName)
+                GameServerEnum.fromPackageName(foregroundAppName)?.let { server ->
```
reconman, line 286: Just for clarity, there should be no case where `fromPackageName` returns null after `onServiceConnected()` was called, right?

But we probably can't replace the ? with a ! because onAccessibilityEvent could be called at the same time as onServiceConnected.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
```diff
@@ -230,28 +253,40 @@ class ScriptRunnerService : AccessibilityService() {
 
     override fun onInterrupt() {}
 
+    private fun isLandscape() =
+        userInterface.metrics.let { it.widthPixels >= it.heightPixels }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {
```
reconman, line 259: I don't remember, is there an active event filter for this similar to:
```
<activity android:name=".MyActivity"
          android:configChanges="orientation|screenSize|screenLayout|keyboardHidden"
          android:label="@string/app_name">
```

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
MathewSachin: Yes

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
MathewSachin: We should always avoid `!!`. That defeats the purpose of null-safety.

app/src/main/java/com/mathewsachin/fategrandautomata/accessibility/ScriptRunnerService.kt
MathewSachin: That is for Activities. Activities are destroyed and recreated by default on configuration changes. That attribute can be set to tell Android not to destroy an activity on config changes and instead call the handler.
Services are not recreated on config changes so there's no manifest attribute to set for them.

reconman: So what kind of configuration changes are you listening for? All of them?

MathewSachin: doesn't matter if I listen or not, that function is called for all config changes.

reconman (approved):

MathewSachin:
@reconman What should we do about the case of switching to portrait orientation while the script is running?

reconman:
I forgot about the pausing suggestion, I agree with that.

PR #529, opened 2020-10-19
Pause script when switching to portrait orientation

MathewSachin:


reconman:
Sorry for not testing earlier, seems to work fine except for one thing that affects all pause triggers. I'll create an issue in a few minutes.

PR #535, opened 2020-10-25
Create  Use-for-other-games

TekpreXyz:
Use for other games

I would need help simplifying your project.
I currently use ankulua and have working scripts for some games. 
I would like to create a universal base app that allows you to use lua scripts. 
Basically an open version of ankulua. The idea is to have an app with only one button,
for selecting scripts and a folder to load the images to search with openCv.

### DIFF
--- a/Use-for-other-games
+++ b/Use-for-other-games
@@ -1,0 +1,9 @@
+
+Use for other games
+
+
+I would need help simplifying your project.
+I currently use ankulua and have working scripts for some games. 
+I would like to create a universal base app that allows you to use lua scripts. 
+Basically an open version of ankulua. The idea is to have an app with only one button,
+for selecting scripts and a folder to load the images to search with openCv.


minhducsun2002:
This app doesn't use any Lua code, I think?

TekpreXyz:
Maybe it's possible to traslate Lua in Kotlin or embled it ?
https://github.com/dickensas/kotlin-gradle-templates/tree/master/embed-lua

MathewSachin:
First of all, this shouldn't have been a pull-request. It had to be an Issue labelled as a `question`.
So, will be closing this.

> I would need help simplifying your project.

Ok. What is your programming experience?
Like, which languages you know? What kind of things have you made?

> I currently use ankulua and have working scripts for some games.
I would like to create a universal base app that allows you to use lua scripts.

Ah, AnkuLua is based on [Sikuli](https://github.com/RaiMan/SikuliX1) which is also open-source.
But, FGA has a custom library called `libautomata` which is considerably different from Sikuli.
So, the existing AnkuLua scripts won't work directly with it since the API is different.
You're better off starting with Sikuli if you want to run existing scripts.

> Maybe it's possible to traslate Lua in Kotlin or embled it ?
https://github.com/dickensas/kotlin-gradle-templates/tree/master/embed-lua

I don't know. Maybe that or something else could be used. I personally just don't like Lua as it feels dated.
But, I doubt `libautomata` would work properly with Lua since it doesn't support things like interfaces having extension functions which I've made heavy use of in the code.

> and a folder to load the images to search with openCv.

This is going to get troublesome with Android 11. Read up on scoped storage.

PR #537, opened 2020-10-25
Script Exit handled outside EntryPoint

### LINKED ISSUE #536: AP ran out does not stop the script in the right order
**Describe the bug**
With the new pause triggers for screen turned off, I discovered one issue with old code.

If AP runs out, the script is stopped only AFTER you click on OK. If the screen is locked while the AP message is on screen, the message about the script being paused appears on top of that.

So I think the order of the messages and the actual action (stop) need to be switched. The order for pausing seems to be correct, since there's no pause button anymore.

**Screenshots**

MathewSachin:
- Added a Stopping state to script states.
- fixes #536
- Message boxes and notifications are not shown by EntryPoint, but by ScriptManager.

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/ui/support_img_namer/SupportImageNamer.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/ui/support_img_namer/SupportImageNamer.kt
@@ -12,6 +12,10 @@ import com.mathewsachin.fategrandautomata.scripts.entrypoints.getFriendImgPath
 import com.mathewsachin.fategrandautomata.scripts.entrypoints.getServantImgPath
 import com.mathewsachin.fategrandautomata.util.dayNightThemed
 import com.mathewsachin.fategrandautomata.util.showOverlayDialog
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.resume
 
 // *, ?, \, |, / are special characters in Regex and need to be escaped using \
 private const val InvalidChars = """<>"\|:\*\?\\\/"""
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/mathewsachin/fategrandautomata/ui/support_img_namer/SupportImageNamer.kt
```diff
@@ -107,17 +111,22 @@ fun showSupportImageNamer(UI: ScriptRunnerUserInterface, storageDirs: StorageDir
...
 
-    showOverlayDialog(context) {
-        setCancelable(false)
-            .setTitle(UI.Service.getString(R.string.support_img_namer_title))
-            .setView(content)
-            .setPositiveButton(UI.Service.getString(R.string.support_img_namer_done)) { dialog, _ ->
-                if (entryList.all { it.isValid() }) {
-                    if (entryList.all { it.rename() }) {
-                        dialog.dismiss()
+    suspendCancellableCoroutine<Unit> { coroutine ->
```
reconman, line 114: I can't wrap my head around this. What does suspendCancellableCoroutine do in this case? I tried reading about it, but the only thing I can find is that suspendCancellableCoroutine is supposed to replace callbacks.

MathewSachin: It turns callbacks into coroutines. The benefit is instead of passing a completion callback to `supportImageNamer` function. I can call it as if I'm calling a sequential function from the main thread, but at the same time not blocking the main thread. And, once the dialog is closed by the user, the rest of the code runs.

reconman: If I understand correctly, the `showOverlayDialog` is executed immediately when this line is reached and then when either the positive or the negative button is hit, `coroutine.resume()` is called.

`coroutine` in this case is the method `showSupportImageNamer`, which has nothing left to do, so the the method which called `showSupportImageNamer` is resumed.

Is that right?

MathewSachin: yes, that's how it works.
Kotlin coroutines are a bit different from async/await in C# and JS in that the default behaviour is to execute code sequentially, unlike C#/JS where you have to `await` for sequential execution.

Issue #599, opened 2020-12-11
TW app update to ver. 1.66.0


the3dsandwich:
Before opening an issue:
- [x] No one has asked for this feature before in other issues (even the closed ones)

> If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

Hi, me again. Later today the client for TW will be updated to version 1.66.0, which from reading the update log should change a few things, most noticeably the addition of the "continue 10 roll" button.

I'm thinking there might be something that needs to be updated when TW is up to 1.66.0, so I'll be on the lookout if that's the case.

the3dsandwich:
So after looking around, the continue roll is the only thing that needs updating. 

I got the image here, and I tried updating the script for it now. If the PR doesn't break anything else then please consider merging, or otherwise the image is here and you may add it in as you normally would. 

![image](https://user-images.githubusercontent.com/32169552/102013331-4e196380-3d8a-11eb-94df-67c776864f5b.png)

Thanks and good luck on the Christmas event (that's the current event on NA right?)

MathewSachin:
The PR was fine. Merged it.
Thanks. This Christmas should ease the bone and fang hell a bit 😆.

PR #602, opened 2020-12-13
TW client 1.66.0 support

the3dsandwich:
Adds script and image for continuing 10 rolls for TW as described in #599 now that TW is up to version 1.66.0


MathewSachin:
Looks good! Merging 👍

Issue #609, opened 2020-12-16
Option to wait for AP regeneration instead of exiting when AP ran out


the3dsandwich:
Before opening an issue:
- [x] No one has asked for this feature before in other issues (even the closed ones)

> If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

Sometimes I might want to take a long (11+hr) sleep but I don't want AP from filling all the way up, so I think an option that makes FGA continue to check if AP has regenerated will be nice.

So far I made some adjustments and tested and it goes as the following:

![Screenshot_20201216-112042](https://user-images.githubusercontent.com/32169552/102301093-20523b80-3f91-11eb-8781-bb438d251019.png)

1. I keep my phone plugged in all night (and all day since it's a spare phone and can be left in the same spot for days)
2. Turn on the option to wait for AP regeneration and turn off Refill
3. When in AutoBattle, after the refill message pops up after clicking on a node, FGA clicks close and waits for 20 seconds (I think some phones have a 30 second screen timeout, I'm not sure if FGA can keep the phone awake without clicking for a long time so I set it to 20 seconds for now)
4. After 20 seconds FGA clicks on the node again to check if AP has regenerated enough to rerun the node

This might also help people that don't want to commit all apples from the beginning in Christmas 2020 in JP but if JP UI update takes more time to adjust to that was more important than this update.

Anyways, I'll make a PR and please see if this sounds good.

ScathachSkadi:
I thought this feature wasn't getting implemented... Anyway it's good to finally get it. Does this work for all servers yet?

MathewSachin:
Yeah, I was against it, but, anyways since @the3dsandwich sent a PR and it's toggleable and OFF by default, I think it's fine.
It should work on all servers since it's nothing specific.

PR #610, opened 2020-12-16
WIP: Wait for AP regeneration

the3dsandwich:
As mentioned in #609 this should add a option in app preference that changes `pref.waitAPRegen`, which is detected in AutoBattle.kt to keep FGA running to keep checking for AP regeneration.


MathewSachin:
A lot of files not related to this feature are changing which is why the build is failing.

You should start by creating a new branch before sending PRs.
That way your fork's master branch can be kept in sync with mine when your PR is merged.
I think it may be easier to fork again instead of resolving conflicts this time.

I wasn't planning to add this feature due to the potential battery hog due to keeping the screen awake for so long, but, I guess it's fine as a toggleable feature which is OFF by default.

the3dsandwich:
Right. Let me see if I can make the changes work for the CI build. Just a sec.

PR #624, opened 2020-12-21
Enable x86 and x64

### LINKED ISSUE #623: Crash on Bluestacks 64-bit
Before opening an issue:

**Describe the bug**
FGA crashes on Bluestacks 64-bit

**Screenshots**
image

The issue here seems to bit libhoudini (ARM to x86/64 translation) isn't working well with the app.

Proposed solution is to add x86 and x86_64 to the ndk > abiFilters section in gradle. I have tested and confirmed it's working.

 - Device model: Bluestacks 64-bit

thai-ng:
Enable x86 and x64 ABI to prevent crash on emulator.
Fixes #623


MathewSachin:
It doesn't crash on my Bluestacks. To confirm I installed the latest Bluestacks 64-bit for Windows and downloaded and ran the Canary build on it.
Bluestacks creates a 32-bit Nougat instance by default which runs both FGO and FGA just fine.

But then, adding more ABIs increases the size of the APK heavily. It went from 32.8MB to 123MB which is quite significant, which is more relevant since we're not on PlayStore for now.
I'll suggest that you create 32-bit Nougat instance in BlueStack's Multi-Instance Manager. 123MB is too much for every build.

thai-ng:
Would it be better to introduce a split apk config then? building out an apk for each ABI

MathewSachin:
I don't think so. It's kind of hard to assume that Android users would normally know about ABIs.

thai-ng:
Up to you then. The reason why I'm running 64-bit Bluestacks is because I was playing another game (Priconne) that only runs on 64 bit Android. I imagine eventually there will be more players playing games with that requirement and also play FGO.

minhducsun2002:
> Would it be better to introduce a split apk config then? building out an apk for each ABI

In line with the author's reply, I doubt there's any simple method to install split APKs. Does BlueStacks support them out of the box, or we need to install `adb` for this purpose?

thai-ng:
> EDIT : Split APK is a concept mentioned [here](https://developer.android.com/guide/app-bundle#split_apks). If you meant an APK for each ABI, please ignore this comment. Apologize for any misunderstandings caused.

Yeah I'm talking about 1 apk for each ABI, not app bundle.

MathewSachin:
#748 and #740 reported errors about the app not working on Nox and Bluestacks.
So, I confirmed that on Bluestacks 5, the latest canary crashed on clicking on the PLAY button.
I thought it might be related to this PR, but, sadly a build from this PR also crashes the same way.

reconman:
Obsolete now that the app is published again on the play store.

PR #634, opened 2020-12-23
Remove some compiler and lint warnings

reconman:
There are still 2 compiler warnings left:
https://github.com/Fate-Grand-Automata/FGA/pull/634/checks?check_run_id=1602420740#step:5:133

And the one about using deprecated features is probably because of the safeargs plugin, which will be updated some time in the future.


MathewSachin:
I was withholding this to not cause conflicts for the translation PRs. They'll have more trouble resolving things. CN one is still left.

Also, a few more things changed. #643 added back Game Server option, so the `ic_earth` icon and the translation strings for it are now being used.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/MediaProjectionScreenshotService.kt
```diff
@@ -34,7 +34,7 @@ class MediaProjectionScreenshotService(
         val screenWidth = DisplayMetrics.widthPixels
         val screenHeight = DisplayMetrics.heightPixels
 
-        imageReader = ImageReader.newInstance(screenWidth, screenHeight, PixelFormat.RGBA_8888, 2)
+        imageReader = ImageReader.newInstance(screenWidth, screenHeight, ImageFormat.FLEX_RGBA_8888, 2)
```
MathewSachin, line 37: I don't see a deprecation notice for this field. Was there a Lint warning?

MathewSachin: The online documentation mentions some other fields as deprecated but not this one. Can't check on my laptop right now.

reconman: The Android documentation is contradicting itself. https://developer.android.com/reference/android/media/ImageReader#newInstance(int,%20int,%20int,%20int) says that `PixelFormat` constants are okay, but then only lists a bunch of `ImageFormat` constants as possible values.

Android Studio uses that ImageFormat list for linting.

MathewSachin: Ok. Got it

reconman:
I still need some help with the `setTargetFragment` warnings. Do you have an idea how the new code should look like based on the deprecation comment found [here](https://developer.android.com/reference/androidx/fragment/app/Fragment?hl=en#setTargetFragment\(androidx.fragment.app.Fragment,%20int\))?

MathewSachin:
I think we don't need to worry about every deprecation warning. I wrote that code based on what the Android team was using in Jetpack Preferences.

reconman:
You're right, the preference library still uses `getTargetFragment()`, which can only return non-null values after calling the deprecated `setTargetFragment`.

MathewSachin (approved):

reconman:
Don't merge yet, it crashes on my phone right now.

MathewSachin:
Well, I couldn't run and check since we're having a power cut here and my laptop is down 😆

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/MediaProjectionScreenshotService.kt
reconman: PixelFormat is the correct one, since ImageFormat doesn't work on my machine because of this method:

https://chromium.googlesource.com/android_tools/+/refs/heads/master/sdk/sources/android-25/android/media/ImageUtils.java#43

It says to use PixelFormat for screenshots and ImageFormat for image files.

app/src/main/java/com/mathewsachin/fategrandautomata/ui/skill_maker/SkillMakerActivity.kt
```diff
@@ -13,8 +13,11 @@ class SkillMakerActivity : AppCompatActivity(R.layout.skill_maker) {
     val vm: SkillMakerViewModel by viewModels()
     val args: SkillMakerActivityArgs by navArgs()
 
-    fun isMainScreen() = findNavController(R.id.nav_host_fragment_skill_maker)
-        .currentDestination?.id == R.id.main_skill_maker
+    val navHostFragment by lazy {
```
reconman, line 16: The old code crashes the app after the change from fragment to FragmentContainerView and Google tells everyone to use the new code in https://issuetracker.google.com/issues/142847973#comment15

app/src/main/java/com/mathewsachin/fategrandautomata/ui/MainActivity.kt
```diff
@@ -33,13 +36,12 @@ class MainActivity : AppCompatActivity() {
         setContentView(binding.root)
         setSupportActionBar(binding.toolbar)
 
-        val navController: NavController = findNavController(R.id.nav_host_fragment)
+
```
MathewSachin, line 39:
```suggestion
```

just an empty line

app/src/main/java/com/mathewsachin/fategrandautomata/ui/skill_maker/SkillMakerViewModel.kt
```diff
@@ -46,6 +46,9 @@ class SkillMakerViewModel @ViewModelInject constructor(
                             m.skillCommand.removeLast()
                             m.skillCommand.add(SkillMakerEntry.Next.Wave(l.action))
                         }
+                        else -> {
+                            //do nothing
```
MathewSachin, line 50:
```suggestion
                            // do nothing
```

just a nitpick

MathewSachin (approved):

app/src/main/java/com/mathewsachin/fategrandautomata/ui/skill_maker/SkillMakerActivity.kt
MathewSachin: Seems to be working just fine.

Issue #633, opened 2020-12-23
Missing translations


reconman:
Currently, there are a bunch of untranslated Strings in the app:

```
"service_runner_description" is not translated in "ko" (Korean), "zh" (Chinese)
"support_imgs_extract_failed" is not translated in "ko" (Korean), "zh" (Chinese)
"skill_maker_space_ishtar" is not translated in "ko" (Korean), "zh" (Chinese)
"skill_maker_emiya" is not translated in "ko" (Korean), "zh" (Chinese)
"skill_maker_quick" is not translated in "ko" (Korean), "zh" (Chinese)
"skill_maker_arts" is not translated in "ko" (Korean), "zh" (Chinese)
"skill_maker_buster" is not translated in "ko" (Korean), "zh" (Chinese)
"start_service" is not translated in "ko" (Korean), "zh" (Chinese)
"stop_service" is not translated in "ko" (Korean), "zh" (Chinese)
"start_service_toast" is not translated in "ko" (Korean), "zh" (Chinese)
"p_refill_none" is not translated in "ko" (Korean), "zh" (Chinese)
"p_spam_summary" is not translated in "ko" (Korean), "zh" (Chinese)
"p_spam_skill" is not translated in "ko" (Korean), "zh" (Chinese)
"p_screenshot_drops_summary" is not translated in "ko" (Korean) or "ja" (Japanese)
"p_wait_ap_regen_text" is not translated in "ko" (Korean) or "ja" (Japanese)
"p_wait_ap_regen_text_summary" is not translated in "ko" (Korean) or "ja" (Japanese)
"p_storage" is not translated in "ko" (Korean), "zh" (Chinese)
"p_folder" is not translated in "ko" (Korean), "zh" (Chinese)
"p_choose_folder_title" is not translated in "ko" (Korean), "zh" (Chinese)
"p_choose_folder_message" is not translated in "ko" (Korean), "zh" (Chinese)
"p_choose_folder_not_exist_title" is not translated in "ko" (Korean), "zh" (Chinese)
"p_choose_folder_not_exist_message" is not translated in "ko" (Korean), "zh" (Chinese)
"p_choose_folder_action" is not translated in "ko" (Korean), "zh" (Chinese)
"p_ignore_notch_summary" is not translated in "ko" (Korean), "zh" (Chinese)
"p_record_screen_summary" is not translated in "ko" (Korean), "zh" (Chinese)
[...]

PR #635, opened 2020-12-24
TW translation

the3dsandwich:
As mentioned in #633, adds in translation for new functionalities and fixes some old translation
It's still work in progress, but I would like to ask a couple questions regarding the new functions:

1. Does **Swipe Multiplier** multiply the duration or the distance being swiped?
2. **Wait before Cards** sets the waiting time before clicking on the attack button (to select command cards), right?
3. Where are `p_choose_folder_title` (**Choose Folder**) and the other strings in the block used?
4. Materials. Does FGA now track how many materials are farmed for FQs? Or what kind of feature is it?

I still have to finish translating the materials and do an overall translation check, but just putting here so people know this is being worked on.


the3dsandwich:
I think the translation is accurate enough for a review.

I changed `app_preferences.xml` so the **Storage** string can be translated, although I admit it is a bit unrelated to TW translation but more of a in-general translation fix.

Just a side note, I feel like some new features are not clearly descripted enough in-app and sometimes I have the urge to add a summary text for the options, because while some  features is self-explanatory in English it may need more description when translated. I was thinking if we can either add some summary text to each option, or somehow put a change log to somewhere (in-app or a .md file on the repo) so people can know what has changed between updates.

MathewSachin:
> Does Swipe Multiplier multiply the duration or the distance being swiped?

distance being swiped

> Wait before Cards sets the waiting time before clicking on the attack button (to select command cards), right?

It's the time waited after clicking on the Attack button and before the cards (face-cards, NP) themselves show up.
It's useful for slower phones, where the Face-cards, NP take longer to show up.

> Where are p_choose_folder_title (Choose Folder) and the other strings in the block used?

The app now asks the user to pick the folder to use for storing support images instead of having full access to file-system. This was done to make it ready to target Android 11. Those strings are used in the dialog boxes shown to ask user to pick a folder.

> Materials. Does FGA now track how many materials are farmed for FQs? Or what kind of feature is it?

Yes, FGA tracks the number of materials farmed of the types you specify.

> I changed app_preferences.xml so the Storage string can be translated, although I admit it is a bit unrelated to TW translation but more of a in-general translation fix.

ah, yeah, that was a miss from my side. Thanks for fixing 👍 

> Just a side note, I feel like some new features are not clearly descripted enough in-app and sometimes I have the urge to add a summary text for the options, because while some features is self-explanatory in English it may need more description when translated. I was thinking if we can either add some summary text to each option, or somehow put a change log to somewhere (in-app or a .md file on the repo) so people can know what has changed between updates.

I used to write changelogs before on GitHub Releases from where the app was downloadable before it was put on PlayStore and I still got questions on the same things written there. i.e. people weren't reading those, so I eventually lost interest in making changelogs.
I did plan to add more summary stuff in the app and make it more beginner-friendly, but with all the toxicity from the community, I've decided to spend my time on features I personally care about. You're welcome to add summary texts and I can let you know what a feature does if you're not sure about it 👍.

I can't read Chinese, so the only thing I'm looking at is the syntax which looks fine to me 👍

MathewSachin (approved):

the3dsandwich:
Thanks for the information! I changed the swipe multiplier text to reflect the distance swiped. If you want to then this PR is ready to merge.

I'm sorry people just don't do their research before using and are downright toxic. I wish I could tell you to ignore them, but I also understand there's no way they will go away on their own, so if there is anything we can help I'm sure the contributors here are happy to help.

Take care and Merry Christmas (or the day after)!

PR #636, opened 2020-12-24
KR translation update

ScathachSkadi:
Uh... I couldn't make my fork up to date, but I did change the localized.xml. The translations should be correct all except for the last few lines which I am not sure of.

<string name="p_shuffle_cards_when_no_effective">No effective cards</string>
    <string name="p_shuffle_cards_when_no_np_matching">No cards matching NP</string>

What does these two do exactly?


ScathachSkadi:
it has conflicts probably because I messed up some of the lines...

the3dsandwich:
From my understanding, when choosing command cards for an attack, FGA can now check if there are EFFECTIVE cards (not RESIST) and/or if there are cards that are the same as the servant using an NP. If there are none, then FGA can go back and shuffle cards.

MathewSachin:
I'll try to resolve the conflicts 👍

MathewSachin:
@ScathachSkadi I've resolved the conflicts. Let me know when you're done with the translation.

To avoid further conflicts, you should ideally create a new branch when creating a pull-request so that the master branch can stay in sync even after your changes are accepted. This might sound a bit confusing though.

ScathachSkadi:
@MathewSachin It's ready now. I checked the app and it looks fine.

PR #637, opened 2020-12-25
Add CN translate

Lineeee217:
Add CN translate

### DIFF
--- a/app/src/main/res/values-b+zh+CN/localized.xml
+++ b/app/src/main/res/values-b+zh+CN/localized.xml
@@ -456,7 +456,7 @@
 
     <string name="p_storage">储存</string>
 
-    <string name="p_folder">Change folder used for Support images, drops images, etc.</string>
+    <string name="p_folder">改变图片储存文件夹.</string>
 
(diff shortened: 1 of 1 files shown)


the3dsandwich:
Hey @Lineeee217 I think for translating the materials, you can check out the [Mooncell page](https://fgo.wiki/w/%E9%81%93%E5%85%B7%E4%B8%80%E8%A7%88). They have the translation for CN materials, if you have trouble finding the translations.

reconman (changes requested):
Can you please use the same format of line breaks as the other files? I don't really like all the empty lines.

Did you use some translation tool which added all the line breaks?

the3dsandwich:
To be honest, I wouldn't recommend using a translation tool to translate from TW to CN since the UI between each of the two clients are different. I have hold off translating for CN personally specifically because I'm not on CN, so I don't know how different they are. 

For example, CN doesn't use "Buster", "Arts" and "Quick" to call the command cards, but instead uses simplified Chinese versions of them. 

So you might want to also check if you are using the correct terms to call each elements.

PR #638, opened 2020-12-26
Restore root functionality for all Android versions

potchy:
Okay, I'm a complete noob when it comes to Kotlin/Android development, so I apologize in advance for code smells, if any.

I hugely prefer root mode over media projection, so I wanted to restore this feature after it has been pretty much killed by 92b9ae71. After a bit of searching, I found that the `screencap` command prints directly to `stdout` if you omit the file parameter (see [here](https://stackoverflow.com/questions/51782364/screencap-directly-tobytearray-skipping-saving-to-memory/51782529) and [here](https://github.com/aosp-mirror/platform_frameworks_base/blob/master/cmds/screencap/screencap.cpp)), so that's exactly what I did. I had to tweak some stuff in `SuperUser` to make `su`'s input stream available to `RootScreenshotService`, so the latter could consume it directly. I feel that those two files are more coupled together now, so I would really appreciate some input with this.

Despite my concerns with the code quality, it's working great, as you can see in the video attached. 😄 

https://user-images.githubusercontent.com/4316326/103143482-b0ad2080-46f5-11eb-8b6f-a20a2af02468.mp4

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
@@ -21,47 +21,37 @@ class RootScreenshotService(
     val storageProvider: StorageProvider,
     val platformImpl: IPlatformImpl
 ) : IScreenshotService, IColorScreenshotProvider {
-    companion object {
-        fun canUseRootForScreenshots() =
-            Build.VERSION.SDK_INT < Build.VERSION_CODES.P
-    }
-
+    private var reader: DataInputStream = SuperUser.inStream
     private var buffer: ByteArray? = null
-    private val imgPath = storageProvider.rootScreenshotFile
 
     private var rootLoadMat: Mat? = null
     private val rootConvertMat = Mat()
     private val pattern = DroidCvPattern(rootConvertMat, false)
 
     private fun screenshotIntoBuffer() {
-        SuperUser.sendCommand("/system/bin/screencap $imgPath")
+        SuperUser.writeLine("/system/bin/screencap")
 
-        imgPath.inputStream().use {
-            DataInputStream(it).use { reader ->
-                val w = reader.readIntLE()
-                val h = reader.readIntLE()
-                val format = reader.readIntLE()
+        val w = reader.readIntLE()
+        val h = reader.readIntLE()
+        val format = reader.readIntLE()
 
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    reader.readIntLE()
-                }
-
-                if (buffer == null) {
-                    // If format is not RGBA, notify
-                    if (format != 1) {
-                        platformImpl.toast("Unexpected raw image format: $format")
-                    }
-
-                    Timber.debug { "${w}x${h} format=$format" }
-
-                    buffer = ByteArray(w * h * 4)
-                    rootLoadMat = Mat(h, w, CvType.CV_8UC4)
-                }
-
-                buffer?.let { b -> reader.read(b, 0, b.size) }
-            }
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            reader.readIntLE()
         }
 
+        if (buffer == null) {
+            // If format is not RGBA, notify
+            if (format != 1) {
+                platformImpl.toast("Unexpected raw image format: $format")
+            }
+
+            Timber.debug { "${w}x${h} format=$format" }
+
+            buffer = ByteArray(w * h * 4)
+            rootLoadMat = Mat(h, w, CvType.CV_8UC4)
+        }
+
+        buffer?.let { b -> reader.read(b, 0, b.size) }
         rootLoadMat?.put(0, 0, buffer)
     }
 
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
@@ -51,7 +51,9 @@ class RootScreenshotService(
             rootLoadMat = Mat(h, w, CvType.CV_8UC4)
         }
 
-        buffer?.let { b -> reader.read(b, 0, b.size) }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
```diff
@@ -21,47 +21,39 @@ class RootScreenshotService(
...
 
-                buffer?.let { b -> reader.read(b, 0, b.size) }
-            }
+            buffer = ByteArray(w * h * 4)
+            rootLoadMat = Mat(h, w, CvType.CV_8UC4)
         }
 
+        // "readFully" will wait for the entire data (b.size) to be available in the input stream,
+        // however long it takes (in actuality, it just takes a few milliseconds).
+        buffer?.let { b -> reader.readFully(b, 0, b.size) }
```
potchy, line 56: I banged my head for like 5 hours total before noticing that I wasn't actually fetching the entire input lmao

app/src/main/java/com/mathewsachin/fategrandautomata/root/SuperUser.kt
```diff
@@ -47,7 +50,7 @@ class SuperUser : AutoCloseable {
     /**
      * Writes a line to the shell.
      */
-    private fun writeLine(Line: String) {
+    fun writeLine(Line: String) {
```
potchy, line 53: `sendCommand` and `waitForCommand` aren't being used anymore.
should we delete them?

app/src/main/java/com/mathewsachin/fategrandautomata/ui/prefs/MoreSettingsFragment.kt
```diff
@@ -57,15 +57,6 @@ class MoreSettingsFragment : PreferenceFragmentCompat() {
 
             it.summary = storageProvider.rootDirName
         }
-
```
potchy, line 60: In this same file, I noticed that the lines 76-84 were also introduced with 92b9ae71, but I'm not sure what they are doing. So I left them there. 😿

app/src/main/java/com/mathewsachin/fategrandautomata/util/StorageProvider.kt
```diff
@@ -45,9 +45,6 @@ class StorageProvider @Inject constructor(
     val rootDirName
         get() = dirRoot?.name
 
-    val rootScreenshotFile
```
potchy, line 48: We're not writing to a file anymore, so it's okay to remove this.

app/src/main/java/com/mathewsachin/fategrandautomata/root/RootScreenshotService.kt
MathewSachin: I had tried reading from the console before too, but gave up after a while when the image won't come of the correct size and resorted to reading from file.
I guess `readFully` was the key 😄.

app/src/main/java/com/mathewsachin/fategrandautomata/root/SuperUser.kt
MathewSachin: Those functions and a common `SuperUser` class were created because I tried out Gestures using root instead of AccessibilityServices ([RootGestures.kt](https://github.com/Fate-Grand-Automata/FGA/blob/b618fe1e042408dda3a3933d1262a6da208818e5/app/src/main/java/com/mathewsachin/fategrandautomata/root/RootGestures.kt)), which would've allowed the app to run on Android 5 and 6. But, the performance wasn't good, so I left it out since every emulator supports Android 7 anyways, and the app can't be backported to Android 5 or 6 due to a lot of APIs we use now.

We don't need those two functions now, so I guess it's fine to remove.

app/src/main/java/com/mathewsachin/fategrandautomata/ui/prefs/MoreSettingsFragment.kt
MathewSachin: `storageProvider.rootDirName` is a bad naming choice from my side. I meant the topmost directory the user picked for storing files here, which can easily be confused with super-user (root).

76-84 are disabling the In-built recording feature when using `Root for Screenshots`.
The in-built recorder uses the same `MediaProjection` used for screenshots to perform the recording, which won't work when using Root, so it is disabled.

MathewSachin:
> Okay, I'm a complete noob when it comes to Kotlin/Android development, so I apologize in advance for code smells, if any.

Hey, the code is absolutely fine, not like I'm an expert in Kotlin/Android 🤣. This is kind of my first Android app.

> I feel that those two files are more coupled together now, so I would really appreciate some input with this.

I think it's fine. You could remove the `sendCommand` and `waitForCommand` now, but it's fine for me either way.

MathewSachin (approved):

potchy:
Great! I've just woke up and I'll do the last changes in a few minutes

app/src/main/java/com/mathewsachin/fategrandautomata/ui/prefs/MoreSettingsFragment.kt
reconman: When I did my youtube guide, I recorded the FGO gameplay using root and recorded my voice using MediaProjection.

So recording with root definitely worked back then.

app/src/main/java/com/mathewsachin/fategrandautomata/root/SuperUser.kt
potchy: I removed `sendCommand`, but kept `waitForCommand` because it contains a piece of knowledge that could be useful to future readers 🤔

app/src/main/java/com/mathewsachin/fategrandautomata/ui/prefs/MoreSettingsFragment.kt
potchy: Oh, I didn't mind `rootDirName`, actually. I wanted to comment lines 76-84, but GitHub limits the scope of the reviews to the parts of the file that were changed only. Thanks for the explanation :)

potchy:
And done!

PR #641, opened 2020-12-28
Added CN Translation

Lineeee217:
Reference to TW translation.
Some idiomatic grammar has been modified.
The translation is consistent with UI.

### DIFF
--- a/app/src/main/res/values-b+zh+CN/localized.xml
+++ b/app/src/main/res/values-b+zh+CN/localized.xml
@@ -1,2 +1,348 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources></resources>
+<resources>
+    <string name="battle_config_item_copy">复制</string>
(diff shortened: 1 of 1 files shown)


the3dsandwich:
Would you mind changing from 巨人的指环 to 巨人的戒指? The former is the TW specific translation and the latter is how CN calls it (Giant's Ring). Although I cannot be 100% sure if that's what it is really called since I don't have CN installed.

Referencing from [Mooncell](https://fgo.wiki/w/%E9%81%93%E5%85%B7%E4%B8%80%E8%A7%88)

Lineeee217:
Thanks. You are right. I will revise it later.

MathewSachin:
I can't say anything about the correctness of the translation since I don't know Chinese, but the formatting is fine this time. I'll leave it open for a while so others can review if they like to. Thank you for translating 👍.

Issue #571, opened 2020-11-25
New JP UI Update - Not working with non-16:9 display


buitim:
Before opening an issue:
- [x] I tested the latest release
- [x] I looked at other issues (even the closed ones)
- [x] I read the [Troubleshooting Guide](https://github.com/MathewSachin/Fate-Grand-Automata/wiki/Troubleshooting)
- [x] I'll fill this template properly

> Your issue will get closed/locked by the bot in case proper info isn't provided
If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

**Describe the bug**
FGO JP just dropped an update where they made their UI adaptive. This removes the blue bars on the sides and messes up the areas that are checked. ~~I personally think they're just out to get you lol /s~~

I haven't tested yet but it might still work on regular 16:9 screens.

**Screenshots**
![image](https://user-images.githubusercontent.com/29244321/100223882-ee3d4480-2ed0-11eb-8473-ad44ad162fdb.png)

**Device Info (please fill at-least 'Device model'):**
 - Device model: Xiaomi Mi 9T
 - Screen size: 1080x2340
 - Android Version: Android 10 - Pixel Experience

MathewSachin:
> I personally think they're just out to get you lol

I don't think so. Full-screen display was something they should've done long ago.

But, I looked at a lot of screenshots and it just looks so unpolished and lazy. A lot of things are out-of-place. Guess they just released before properly testing out stuff.
There is a work in progress mentioned there so hopefully they polish things up.

Even on 16:9 some buttons have moved up a bit, like the quest Repeat button.
https://community.gamepress.gg/t/automatic-farming-app-fate-grand-automata/72155/2020?u=mathewsachin

I'd rather wait for them to get this messed up UI right, rather than adapt to it right now.

reconman:
@MathewSachin Where did you find the screenshots?

MathewSachin:
On Atlas Academy

buitim:
> I don't think so. Full-screen display was something they should've done long ago.

Haha yeah don't worry I was just joking around. Just some sarcasm :)

> But, I looked at a lot of screenshots and it just looks so unpolished and lazy.

Agreed. It's nice to have but some elements definitely feel out of place. Feels like a hack job. I also heard the blue bars come back when using an NP.

observeroftime02:
[...]

Issue #420, opened 2020-09-17
Battle works fine but doesn't select or scroll support


SleepyAhn:
> MAKE SURE YOU TESTED THE LATEST RELEASE.

> Did you look at other issues and the [Troubleshooting Guide](https://github.com/MathewSachin/Fate-Grand-Automata/wiki/Troubleshooting)?

**Describe the bug**
Skills order, NP and battle all works fine
It's just that I can't loop because FGA doesn't select a support
I tried to full screen one or the other or both together and it doesn't select click/scroll support either way
FGA only works in battle if both are not full screen too 

**Screenshots**
It is recommended that you atleast add one screenshot of FGO running on your device.
These are my settings 

![Screenshot_20200917-142823_FateGrand_Automata](https://user-images.githubusercontent.com/71425793/93428655-6ea06180-f8f2-11ea-991b-fd6113aad09e.jpg)
![Screenshot_20200917-142905_FateGrand_Automata](https://user-images.githubusercontent.com/71425793/93428661-6fd18e80-f8f2-11ea-94a2-118b38315557.jpg)

**Video**
I'm the one who clicked on the support myself at the beginning of the video because FGA doesn't

https://www.youtube.com/watch?v=93Tybnk9n3U&ab_channel=SleepyAran

**Device Info (please fill atleast 'Device model'):**
 - Device model:  Samsung Galaxy S10 Lite
 - Screen size: 1080x2400
 - Android Version: Android 10
 - RAM: 8 RAM

MathewSachin:
First line in the issue literally says this: Did you look at other issues and the Troubleshooting Guide?

SleepyAhn:
Hi, sorry. Did I look at the troubleshooting guide? Yeah, that's why I tried to change around the full screen thing like you wrote in there with the Samsung settings and  it does give blue screens at the side but than FGA stops working. The other issues you mainly told them to go read the same guide about the screen detection area / turning on or off the notch detection or download the latest release but I just downloaded it from google play an hour ago so I just assume it's the latest(?). Sorry I'm really stupid

MathewSachin:
I mean the video you uploaded didn't have blue bars, the area is black, so how would I know?

reconman:
Can you post a video of blue bars not working with debug mode?

reihunter360:
BUG/Problem:
[...]

PR #642, opened 2020-12-28
Add betterfgo APK support

neobenedict:
Add modded FGO APK support by adding the two package names.

### DIFF
--- a/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/enums/GameServerEnum.kt
+++ b/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/enums/GameServerEnum.kt
@@ -1,8 +1,8 @@
 package com.mathewsachin.fategrandautomata.scripts.enums
 
 enum class GameServerEnum constructor(vararg val packageNames: String) {
-    En("com.aniplex.fategrandorder.en"),
-    Jp("com.aniplex.fategrandorder"),
+    En("com.aniplex.fategrandorder.en", "io.rayshift.betterfgo.en"),
+    Jp("com.aniplex.fategrandorder", "io.rayshift.betterfgo"),
     Cn("com.bilibili.fatego", "com.bilibili.fatego.sharejoy"),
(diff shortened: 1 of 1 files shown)


the3dsandwich:
Dumb question but what is betterfgo?

I tried to google but can't seem to find immediate results

neobenedict:
> 
> 
> Dumb question but what is betterfgo?
> 
> I tried to google but can't seem to find immediate results

The main purpose of the pull request is this mod is able to hook `UnityEngine.Screen::get_width()` and `UnityEngine.Screen::get_height()` to force the game's UI to 16:9 while keeping the game border-free. It just has a different package name to FGO.

You can find the mod in the Rayshift discord, it's not distributed widely.

the3dsandwich:
I see. This sounds cool. Thanks for the info!

MathewSachin:
> mod is able to hook UnityEngine.Screen::get_width() and UnityEngine.Screen::get_height() to force the game's UI to 16:9 

FGA started looking outside the 16:9 region from 4a8f68e2db97e00e12ebdbf5d9cff665d5a8c070 and right now is at a stage where wide-screen is kind of working (atleast on the few devices and resolutions I checked). See #571.

> Add modded FGO APK support by adding the two package names.

I'm personally against adding modded APK names into the app. I had denied a request before in https://github.com/Fate-Grand-Automata/FGA/issues/420#issuecomment-695158507 and I really sounded like an idiot there because I was mad at the influx of questions at that point.
That was when Max created a fork and I think he added BetterFGO there too.
If I add BetterFGO, I won't be able to deny adding other modded APK names as I did before. I don't want to track all the modded APK names that are gonna pop up here after that. Hope you understand.

------------------------------

As a solution, I'll think about adding back the option to set the Game Server manually, so that modded APK names don't need to be included. I'll keep its default to auto-detect while still allowing the user to change if needed.

MathewSachin:
#643 adds back the Game Server option which can be changed in case you don't want the auto-detection. This can be used regardless of the APK package name.

PR #647, opened 2020-12-29
Wait for AP Regen Enhancements

the3dsandwich:
I think the time between checks (20 seconds) can be longer since AP only regenerate once per 5 minutes, but considering people may set a shorter screen timeout duration, I think giving an option for them to set their own recheck time can be better.

So it now rechecks from every 1 minute to every hour depends on the setting, and it will display a toast saying how long before next check each minute.

I hope this is not too big of a change.


app/src/main/res/xml/app_preferences.xml
```diff
@@ -63,11 +63,27 @@
...
+                app:summary="@string/p_wait_ap_regen_text_summary" />
+
+            <SeekBarPreference
+                android:max="60"
+                app:icon="@drawable/ic_counter"
+                app:key="@string/pref_wait_ap_regen_minutes"
+                app:title="@string/p_wait_ap_regen_minutes_text"
+                app:summary="@string/p_wait_ap_regen_minutes_text_summary"
+                app:min="1"
+                app:defaultValue="5"
```
MathewSachin, line 82: Good QoL overall, but can we make the default 1 min. People usually have around 1-2 min screen timeout. So, I think that should be a better default.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -460,7 +461,10 @@ open class AutoBattle @Inject constructor(
             3.seconds.wait()
         } else if (waitAPRegenPrefs) {
             Location(1300, 1240).click();
-            20.seconds.wait()
+            for (i in 1..waitAPRegenMinutePrefs) {
+                toast("Waiting for AP regeneration. Re-checking in "+ (waitAPRegenMinutePrefs-i+1) + " minutes.")
```
MathewSachin, line 465:
```suggestion
                toast("Waiting for AP regeneration. Re-checking in ${waitAPRegenMinutePrefs - i + 1} minutes.")
```

just a Kotlin feature

app/src/main/res/values/localized.xml
```diff
@@ -159,7 +159,9 @@
     <string name="p_battle_config_cmd_text_summary">Show text-box for editing Skill commands</string>
 
     <string name="p_wait_ap_regen_text">Wait for AP Regeneration</string>
-    <string name="p_wait_ap_regen_text_summary">Keeps FGA running even when AP runs out, and automatically restart node once enough AP has regenerated.</string>
+    <string name="p_wait_ap_regen_text_summary">Keeps FGA running even when AP runs out and automatically restart node once enough AP has regenerated.</string>
+    <string name="p_wait_ap_regen_minutes_text">Minutes Before Rechecking</string>
+    <string name="p_wait_ap_regen_minutes_text_summary">Default: 5. Please set your screen timeout to be longer than the duration.</string>
```
MathewSachin, line 164:
```suggestion
    <string name="p_wait_ap_regen_minutes_text_summary">Please set your screen timeout to be longer than the duration.</string>
```

The default isn't really crucial to know for users unlike fine-tune. So, I think we could omit that from the summary.

MathewSachin: For the translation too

MathewSachin (changes requested):

app/src/main/res/xml/app_preferences.xml
the3dsandwich: Sure! let me change this and the translation too

the3dsandwich:
@MathewSachin changes are all in place and ready to merge

MathewSachin (approved):

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -460,7 +461,10 @@ open class AutoBattle @Inject constructor(
             3.seconds.wait()
         } else if (waitAPRegenPrefs) {
             Location(1300, 1240).click();
-            20.seconds.wait()
+            for (i in 1..waitAPRegenMinutePrefs) {
```
reconman, line 464: I think if you reverse the order, it should be fine:
```suggestion
            for (i in waitAPRegenMinutePrefs..1) {
                toast("Waiting for AP regeneration. Re-checking in $i minutes.")

**Review by reconman — CHANGES_REQUESTED** [review]

File: scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt (hunk 4 above)
Comment (reconman) [reply scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt:464]: Oh and shouldn't this message be translatable?

Comment (the3dsandwich) [reply scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt:464]: Right, it is! To be honest I'm not sure how to put the translatable string here, what should I do?

Comment (MathewSachin) [reply scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt]: Look into `IScriptMessages`. Similar code is used for showing AP running out.

File: app/src/main/res/values/localized.xml [hunk 5]
```diff
@@ -159,7 +159,10 @@
     <string name="p_battle_config_cmd_text_summary">Show text-box for editing Skill commands</string>
 
     <string name="p_wait_ap_regen_text">Wait for AP Regeneration</string>
-    <string name="p_wait_ap_regen_text_summary">Keeps FGA running even when AP runs out, and automatically restart node once enough AP has regenerated.</string>
+    <string name="p_wait_ap_regen_text_summary">Keeps FGA running even when AP runs out and automatically restart node once enough AP has regenerated.</string>
+    <string name="p_wait_ap_regen_minutes_text">Minutes Before Rechecking</string>
+    <string name="p_wait_ap_regen_minutes_text_summary">Please set your screen timeout to be longer than the duration.</string>
+    <string name="wait_ap_regen_toast_message">Waiting for AP regeneration. Rechecking in %d minutes.</string>
```
reconman, line 165:
```suggestion
    <string name="wait_ap_regen_toast_message">Waiting for AP regeneration. Rechecking in %d minute(s).</string>
```

app/src/main/res/values/localized.xml
```diff
@@ -159,7 +159,10 @@
     <string name="p_battle_config_cmd_text_summary">Show text-box for editing Skill commands</string>
 
     <string name="p_wait_ap_regen_text">Wait for AP Regeneration</string>
-    <string name="p_wait_ap_regen_text_summary">Keeps FGA running even when AP runs out, and automatically restart node once enough AP has regenerated.</string>
+    <string name="p_wait_ap_regen_text_summary">Keeps FGA running even when AP runs out and automatically restart node once enough AP has regenerated.</string>
```
reconman, line 162: This seems a bit awkward. How about
```suggestion
    <string name="p_wait_ap_regen_text_summary">Keep FGO active after the AP have run out. The current node will be repeated once enough AP have regenerated naturally.</string>
```

reconman: The TW text probably also needs to be adjusted to be similar, right?

the3dsandwich: I wrote the TW translation based on the behavior instead of a direct translation, so it goes along the lines of "after AP runs out FGA won't exit immediately, but instead will continue to check whether AP has replenished and restart node".  I think it should be ok to leave it as is...?

reconman: I guess that's alright.

reconman (approved):

the3dsandwich:
Thanks a lot!

PR #650, opened 2020-12-30
Wait AP Regen Fix

the3dsandwich:
Sorry for bothering. Last night thought `..` works for decreasing for loops without checking whether it actually works, and now FGA skips the waiting time and keeps checking nonstop. [Stackoverflow](https://stackoverflow.com/questions/9562605/in-kotlin-can-i-create-a-range-that-counts-backwards) says I should have used `downTo` instead of `..` in a decreasing variable situation so I made a simple fix.


the3dsandwich:
I really should have checked everything before letting it merge...

MathewSachin (approved):
Ah, I had forgotten  `..` doesn't work for reversed loops.
Well, I messed up Stage counter code multiple times yesterday too 😅

the3dsandwich:
Life is a never ending loop of introducing more bugs by fixing them 😂

PR #675, opened 2021-01-28
Jetpack Compose

### LINKED ISSUE #552: Accessibility kill option
I appreciate your contribution to fgo scripting community.
My suggestion is not fixing accessibility problem. I want to have option to forcefully close accessibility after exiting application. I have couple of example applications does that like auto clicker. 
After exiting app, application closes accessibility for itself. It would turn itself off.
 Right now, I need to force stop the app so it can turn off accessibility settings for next time. 
Currently, App gets stucks at accessibility on settings. 
Phone: Xiaomi mi 9T, miui optimisation is off.

### LINKED ISSUE #563: [NA] Lotto mode Reset Prize when out of lotto currency
So it's currently the Christmas Rerun for NA and I notice that when I run the Lotto script to open Lotto box, whenever it runs out of lotto currency and can't open box anymore, it choose reset prize and move to the next box.

I don't think this is intended nor did it happen on previous lotto.
I'm using a Pixel 3, Android 11

### LINKED ISSUE #632: Optional Play/Pause Button Size Override
Would it be possible to add a new setting in fine tune that overrides the auto scaling of the pause play button? I'd like to keep mine small regardless of resolution so that I can keep it in the bottom left corner without the script mistakenly clicking on it. Maybe a slider with small, medium, large? Something to that effect?

Thank you for considering!

### LINKED ISSUE #644: Stop on node completion (first time clear)
Before opening an issue:
- [x] No one has asked for this feature before in other issues (even the closed ones)

- searched 'first clear ' and 'node completion'

> If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

It can be accomplished by having it look at screen for the saint quartz from node completion.

Right now im trying to substitute that by just having it do '3' runs, but it gets messed up if i have to stop the script, meaning i have to go back to change amount of runs or risk not paying enough attention and having it waste AP by repeating a node (especially if first clear has lower AP usage.)

### LINKED ISSUE #686: Continuing to Skill Spam with new Servants coming in
Before opening an issue:

I wanted to ask if it would be possible to add an option for skill spam where after one of the first three Servants die, the remaining servant continues to skill spam.

I am aware that if any skill needs to be targeted it may fail, and/or if a Servant who normally needs to target it might select the servant instead.

That said it'd still overall be more useful than the current approach of new Servants not using there skills at all.

### LINKED ISSUE #733: Auto CE Bomb Script
A script to minimize the amount of tedious work involved in creating CE bombs.

I've made a very basic version of such a script for use with AnkuLua and Grand Order Lua. My script is not perfectly optimal QP wise, but I think it strikes a good balance between QP efficiency, amount of time to run, and difficulty of programming. 

I built my script in AnkuLua half because I know the tool better and half because I couldn't manage to get anything to build and function in Android Studio. My code is mess and probably has a bunch of unnecessary waits and hard coded values for my device, but since it barely functions I don't know that it's very likely I'd improve it anytime soon myself. Since now is a good time for CE bomb making, given the increased success rate and half qp cost, I thought I'd post here in case anyone was interested in working it into FGA or simply using it as is. I saw that there were a few closed issues asking for the same thing, but I thought that since I had work to show it might warrant opening a new issue.

What my script currently does:
 
- Selects a level 1 Starvation (1 Star) or Gloom (2 Star) as the base CE
- Scrolls to the bottom of the CE material selection and clicks on every CE starting from the bottom
- Scrolls to the top of the CE material selection and clicks on a single level 1 of CE of the same kind as the base
- Combines them
- Repeats from step 1 until all CEs have been combined together (or more likely it crashes)

What my script assumes you already did:

- Set both filters to show only 1 star OR 2 star CEs
- Set the display scale to the smallest setting
- Locked anything you don't want lost
- Started the script while sitting on an empty "Tap to select a Craft Essence to enhance" screen.

What my script fails to do:

- When it gets close to the end the 2nd copy of the base CE is sometimes already selected and gets deselected instead. This is sometimes a good thing as it insures having a Lv. 1 CE for the next round, but it can also make it take one round too many.
- Stop elegantly. Currently it runs until it doesn't find a Lv. 1 CE and crashes.

To do perfectly optimal CE bombs you should always be using a 1 star CE as a base. My script does not do this, instead consolidating all 1 star CEs together and all 2 stars CEs together. I have a couple reasons for this.

I have never run into a situation where I didn't have enough copies of Starvation or Gloom to consolidate everything. If instead I always used a 1 star as the base I don't think there would be enough copies of Starvation to do everything, requiring programming in more 1 star CEs and a way to make sure that the base and the 2nd copy are the same CE.
Secondly, a 1 star CE base can only hold six 2 star CEs while a 2 star base hand hold 19 (20, if you count the copy of itself), reducing the amount of times the script would need to run to finish.

Additional things that could improve the script:

- The script setting the filters itself rather than assuming their already set. (Open the filter, cancel all, turn on "not selected" and the desired star rarity)
- The script checking if the scale setting is right. (There could be a problem where the region in question is overlapped by the FGA button by default on some systems.)
- Optimizing the 2nd copy selection

The attached zip contains the lua script and the three used pngs.
[Auto CE Bomb.zip](https://github.com/Fate-Grand-Automata/FGA/files/6678422/Auto.CE.Bomb.zip)

### LINKED ISSUE #736: Servant Death by NP (Arash, Chen Gong, etc)
Before opening an issue:
- &#x2612; No one has asked for this feature before in other issues (even the closed ones)

> If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

I know this topic is discussed in many other issues (all closed now and I probably shouldn't be replying there) but I think I have a suggestion that might work and would like to ask for your opinion.

Suggested approach: **Detect servant death by checking when there are exactly 2/3 sets of the servant HP/NP bar**, and start tapping.

(I know there is a list of the existing issues that asked about the same problem)
#438 #437 #393 #345 #185

So far on those issues we see suggest the following approaches:

- Wait for X seconds then click for 300ms, X is configurable (#345): doesn't work because servants don't die every turn and there needs to be a way to detect servant death
- Add "death" option in autoskill, then script knows when to expect a servant death (#437): doesn't work because we still don't know how long the NP lasts and we also don't know how to detect a servant death.

I think I have a way to detect a servant death, or rather, the timing when we should start tapping.

https://www.youtube.com/watch?v=jzxE6adbgxg

Here's a run of using Chen Gong where each wave a servant dies to illustrate. 

- Let's say there are 3 servants on the field before Chen Gong's NP and there is at least one servant behind.
- After Chen Gong's NP, instead of trying to detect the servant sprite dying, we can instead match/look for the presence of the HP/NP bar (or the LV text underneath). If we see exactly 2 occurrences on the screen, then we know one of them was used as sacrifice and the animation has not finished yet: **we start our tap here**
- We also know that when the animation finishes, there will be again 3 occurrences of the HP/NP bar, plus there will be the attack button.
- If there are not enough servants to replace, then it won't be a problem either since when we check the occurrences of the HP/NP bars we also check whether the attack button is present.

I meant the red circle when I say the HP/NP bar.

Untitled

So I suggest the logic be a combination of the previous suggestions:

1. Add "death" option in autoskill so FGA knows when it should anticipate and detect servant deaths.
2. After an NP, if the "death" option is enabled, FGA waits for a configurable X seconds before matching images on the screen.
3. FGA checks 4 regions: servant 1's HP/NP bar (a), servant 2's HP/NP bar (b), servant 3's HP/NP bar (c), the attack button (d).
4. if d matches, next turn/wave. No taps.
5. if the number of matches for a,b,c is between 1 and 2, initiate taps.

Mikon Anticipate Her Sacrifice While Master Test Out FGA - YouTube and 6 more pages - Personal - Microsoft​ Edge 2021-06-20 14-08-06_1

I wonder if we are able to switch to a different image matching entry point when it's controlled by autoskill. Maybe we might even skip the autoskill check and instead match and count occurrences directly. I cannot be sure. But if this works then it'll be so nice. Thanks again!!!!

### LINKED ISSUE #757: Option to Move to Gift Box Script from the Lottery Script, And Stop Only When Inventory is Full
Before opening an issue:
- [x] No one has asked for this feature before in other issues (even the closed ones)

> If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

Since TW and JP both have lottery events I think some people may have some enhancement ideas, so I would like to pitch in some of my ideas when I was opening my lottery boxes:

- #476 Mentioned combining Lottery/Gift Box scripts together. While combining those scripts entirely is not practical, I find myself wanting the Lottery script to click on the "move to gift box" button (and even want it to just start clicking). Judging from the Jetpack Compose UI, could we add a toggle to say if we want to the gift box script to run right after lottery, and link those two together?
- Gift Box script now selects up to 99 exp stacks and exits. Is there a way we can make it click on the receive button then pick it again, and only stop when FGO ask us to sell exps or move to servant enhancement (or even click on the burn button to go to the shop)?

Basically,
- adding an option to click on the "move to gift box" button and initiate gift box script, and
- making the gift box script receive the selected exp stacks until inventory is full
should make our box opening experience even better.

Let me know if there's any technical difficulties and how I can help/test it, thanks!

MathewSachin:
Jetpack compose is now available in Android Studio Beta and hit Release Candidate.

This is almost a complete rewrite/redesign of the app UI.

## Adjusting for how FGO JP handles notch
`Ignore notch detection` is implied on JP since FGO JP extends over the notch area now.
The PLAY button only shows PAUSE option while running. STOP button comes up after pausing.
The button size (in dp) is now constant across resolutions. (fixes #632)

## Lottery script can start receiving embers when Present box is full
When starting the Lottery script, there is an option to receive embers (fixes #757). You can also configure the gift box script in place.
GiftBox script now continues to receive embers multiple times unless inventory full.
If within a 2s period a connection failure happens, it can retry. If accepting embers takes more than 2s, the script will stop.
This script needs a decent device and a decent internet connection for proper functionality.

## Option to prevent lottery box reset
This helps if you're in the first 10 boxes and don't want FGA to reset the box if you run out of lottery currency (fixes #563).
This shows up in the dialog when starting the script on the lottery screen.

## Option to control the number of FP rolls
This shows up in the dialog when starting the script on the FP screen.

## Option to stop on first-time clear rewards
When you want to collect SQ, apples, etc. from free quests and stop (fixes #644)

## Skill/NP spam configurable for every servant
Spam settings for all 6 servants can be specified individually (fixes #686).

## Battle now skips servant death animation
The most relevant use case is to skip servant death by NP (Arash, Chen Gong, etc) (fixes #736)

## Separate AccessibilityService and ForegroundService
The AccessibilityService was separated out from the foreground service which runs scripts.
Also, allows disabling the AccessibilityService from within the app UI (not exactly what was requested in #552 but I consider it adequate, fixes #552).
Disabling the AccessibilityService before swiping away/closing the app would prevent Accessibility issues.

## Experimental script to create CE bombs
It's there, but READ THE INSTRUCTIONS  (fixes #733)

- Not calibrated for FGO JP.
- Can only be started from CE enhancement screen with no CE selected.
- In the CE picking screens, the item sizes must be set to lowest.
- Base CE pickup screen should be filtered to correct rarity and sorted in Ascending order by Level.
- Enhancement material pickup screen should be filtered to correct rarity and sorted in Descending order by Level.


reconman:
I cherry-picked my commit to this branch so I can try out the new mats when I start farming for Arjuna Alter.

the3dsandwich:
Hey I was wondering if there is a rundown of all the new features this update will bring and perhaps when we'll push this to the main branch?

If not yet then maybe I can help compile a list of the new features I saw so we can all be excited of the update:

- [UI] UI redesign by Jetpack Compose
- [UI] More tabbed views for each menu
- [UI] More icons and images for materials, servant class, etc.
- [UI] Option to prioritize a specific servant's cards (#730)
- [UI] Option to stop on first time clear (#644)
- [Script] New Script to create CE bomb (#733)
- [Battle] More options for skill/NP spam for each servant
- [Battle] Battle now skips servant death animation (#737)

I might miss some things, please let me know if there are more things I can get excited about and maybe help test/look out for. Really excited when this update merges to the main branch!

MathewSachin:
>Hey I was wondering if there is a rundown of all the new features this update will bring and perhaps when we'll push this to the main branch?

I was procrastinating till we actually merge 😅 

> If not yet then maybe I can help compile a list of the new features I saw so we can all be excited of the update:

🎉

> [UI] UI redesign by Jetpack Compose

Most of the new features are UI changes. Writing them all down would be tedious, so we probably should leave it at that 😆.

Another thing I found by experience is that most of the Accessibility issues people face on phones with deadly battery optimizations could be solved by just turning OFF the AccessibilityService before exiting the app. #552 asked for automatically stopping the service, but I currently don't have a way to do that, so instead, I added a button to STOP accessibility service within the app. Starting the AccessibilityService again would require going to phone settings (or not in a case like mine if your device supports turning ON AccessibilityServices with gestures).

> [UI] Option to prioritize a specific servant's cards (#730)

I have implemented the tracking mechanism, but, it's NOT available as a feature in the UI yet.

> [Script] New Script to create CE bomb (#733)

I would rather not advertise this, since it is NOT something that can be used without reading the instructions 🤣.

> Really excited when this update merges to the main branch!

We can merge when Jetpack Compose hits stable which is stated to happen in July.

MathewSachin:
Compose is now in RC. It's about time to merge this PR. :shipit:

PR #737, opened 2021-06-20
Skip Servant Death by NP (Arash, Chen Gong, etc)

the3dsandwich:
As discussed in #736, here's the update. I haven't tested it with enough stages so I fear we might encounter some stages where image recognition may miss the level image (ended up using the level text instead of the HP/NP bar text). I would really appreciate it if people can test it with different stages.

Things this feature probably needs:
- The detecting region needs to be tweaked for wide screen (I'm not on JP so I'm not sure how to set them.
- Now the default location for the overlay button will cover the image recognition for the leftmost servant's level text.
- I didn't create a separate function for clicking on the center like other screen actions does.

What this update does:
- FGA checks 4 regions: servant 1's level text (a), servant 2's level text (b), servant 3's level text (c), the attack button (d).
- if d matches, next turn/wave. No taps.
- if the number of matches for a,b,c is between 1 and 2, make one tap at the center.

Since we know that the center tap only occurs during an animation, it should only skip servant deaths and have no negative impact otherwise.

Video of a run with Chen Gong: https://youtu.be/CDJXHeeXAwA

(will come back tomorrow as it's getting late here)


MathewSachin:
The first thing that I noted from the video is that the `Lv` image isn't matching correctly (easier to see starting from the 2nd wave).
It should be because of the transparent background which would be different for all 3 servants and will also vary depending on the node.
I think the red square I marked in the image below would be a safer option.

image

> Now the default location for the overlay button will cover the image recognition for the leftmost servant's level text.

My above suggestion should handle this.

> I didn't create a separate function for clicking on the center like other screen actions does.

I think it's fine since there's only a single click to do.

> The detecting region needs to be tweaked for wide-screen (I'm not on JP so I'm not sure how to set them.

No worries, I can tweak that later. Let's just get it working on other servers first 😄.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -175,7 +175,9 @@ open class AutoBattle @Inject constructor(
             { needsToStorySkip() } to { skipStory() },
             { isFriendRequestScreen() } to { skipFriendRequestScreen() },
             { isBond10CEReward() } to { bond10CEReward() },
-            { isCeRewardDetails() } to { ceRewardDetails() }
+            { isCeRewardDetails() } to { ceRewardDetails() },
+            { isDeathAnimation() } to {
+                game.scriptArea.center.click() }
```
MathewSachin, line 180:
```suggestion
            { isDeathAnimation() } to { game.scriptArea.center.click() }
```

the3dsandwich:
Got it. I'll be working on it!

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -269,6 +271,11 @@ open class AutoBattle @Inject constructor(
     private fun isCeRewardDetails() =
         images[Images.CEDetails] in game.resultCeRewardDetailsRegion
 
+    private fun isDeathAnimation() =
+        ((if (images[Images.ServantExist] in game.battleServant1Region) 1 else 0) +
+                (if (images[Images.ServantExist] in game.battleServant2Region) 1 else 0) +
+                (if (images[Images.ServantExist] in game.battleServant3Region) 1 else 0)) in 1..2
```
MathewSachin, line 277:
```suggestion
        listOf(
            game.battleServant1Region,
            game.battleServant2Region,
            game.battleServant3Region
        ).count { images[Images.ServantExist] in it } in 1..2
```

MathewSachin: Just trying to simplify the code.

MathewSachin:
> will come back tomorrow as it's getting late here

Sleep is important 😄. We're in no rush.

---

But, whenever you do plan to test it out further:
Right now, there's some part of the background in the image. I meant to crop the region even more. Something like:

image

It is important to avoid that background from the image since it varies per node.

the3dsandwich:
Sure! I'll grab some sleep then make the tweaks 😂

the3dsandwich:
I think this is ready to merge. Updated the image, updated the matching region and tested on some of the nodes. If possible I would prefer matching on more nodes but for now it works on every node I tested.

MathewSachin:
Great! I'll add JP wide-screen support after merging.
I think we should leave the related issue open till the `compose` branch is merged into `master`.

PR #800, opened 2021-08-03
Fix JP FP summon

### LINKED ISSUE #799: Friend gacha auto detection failed after update
**Describe the bug**
Friend gacha script is not loaded even on friend gacha screen after the latest update in JP.
The detection region is shifted as JP added a banner scroll at the bottom of the gacha screen. 

**Screenshots**
mpc-hc64_1FRdwI7O3c

 - Device model:  LD Player 4.0.53
 - Screen size: 1920x1080
 - Android Version: Android 7.1.2
 - RAM: 4GB

reconman:
Fixes #799


MathewSachin:
How about just increasing the region height for all servers. That way we won't have to worry when other servers get this change.

MathewSachin:
`fpFirst10SummonClick` can be made relative to the detected location of `fpSummonCheck` image.

reconman:
I increased the region size as you suggested and changed the 10x summon click position to one that works for all servers.

I chose an X location which is at the top pixel of the "10" in NA and in the middle of the button in JP.

MathewSachin (approved):
works fine on both NA and JP

PR #803, opened 2021-08-04
Include 1 in skill-ten image

### LINKED ISSUE #796: Append Skills from JP breaks the support picker
Before opening an issue:

**Describe the bug**
Support picker no longer working with the Append Skills released for JP's 6th anniversary

**Screenshots** / **Video**
https://imgur.com/a/CSdZv4P

 - Device model: BlueStacks Hyper-V 4.280
 - Screen size: 1920x1080
 - Android Version: Android 7.1.1
 - RAM: 4GB

reconman:
Fixes #796


app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -45,48 +67,71 @@ class DroidCvPattern(
...
-        val result = Mat()
-        resize(result, size)
-        return DroidCvPattern(result).tag(tag)
+        val resizedMat = Mat()
+        resize(mat, resizedMat, size)
+
+        var resizedAlpha: Mat? = null
+        if (alpha != null) {
+            resizedAlpha = Mat()
+            resize(alpha!!, resizedAlpha, size)
```
reconman, line 87: I can't avoid this with a `let`, since the compiler complains that something could have set alpha to `null` between the null check and the method call.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -45,48 +67,71 @@ class DroidCvPattern(
         if (ownsMat) {
             mat?.release()
         }
```
MathewSachin, line 69:
```suggestion
        }
        
        alpha?.release()
```

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -45,48 +67,71 @@ class DroidCvPattern(
         if (ownsMat) {
             mat?.release()
```
MathewSachin, line 68:
```suggestion
            mat.release()
```

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -166,7 +208,7 @@ class DroidCvPattern(
         }
     }
 
-    override fun copy() = DroidCvPattern(mat?.clone()).tag(tag)
+    override fun copy() = DroidCvPattern(mat?.clone(), alpha?.clone()).tag(tag)
```
MathewSachin, line 211:
```suggestion
    override fun copy() = DroidCvPattern(mat.clone(), alpha?.clone()).tag(tag)
```

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -43,45 +65,77 @@ class DroidCvPattern(
...
+                } else {
+                    Imgproc.matchTemplate(
+                        mat,
+                        template.mat,
+                        result,
+                        Imgproc.TM_CCOEFF_NORMED
+                    )
+                }
+
+                // in some cases, NaNs and Infinite values are part of the result
```
MathewSachin, line 130: I don't like this. We should check why we're getting Infinite and NaN in the first place.
These values didn't appear before so something could get bugged.

MathewSachin: This would also make some checks looking at the whole screen much slower 😢

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -43,45 +60,62 @@ class DroidCvPattern(
...
 
     override fun resize(target: IPattern, size: Size) {
         if (target is DroidCvPattern) {
-            resize(target.mat!!, size)
+            resize(mat, target.mat, size)
+            alpha?.let {
+                if (target.alpha == null) {
+                    target.alpha = Mat()
+                }
+                resize(it, target.alpha!!, size)
```
MathewSachin, line 92:
```suggestion
                target.alpha?.let { targetAlpha ->
                    resize(it, targetAlpha, size)
                }
```

I hate `null` exceptions => I hate `!!`.

reconman: A null exception could never happen because I'm changing it to a non-null value in that case.

I don't really like using nested `let`s. How about this instead:
```kt
alpha?.let { originalAlpha ->
    target.alpha = (target.alpha ?: Mat()).apply {
        resize(originalAlpha, this, size)
    }
}
 ```

Comment (MathewSachin) [reply app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt:92]: I'm fine with your suggestion.

> could never happen

you never know how much the code might change in the future.

**Comment by MathewSachin** [comment]
wait! I'm building locally right now. Need more time to test.

File: app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt [hunk 7]
```diff
@@ -17,24 +17,41 @@ import kotlin.math.roundToInt
...
     private val ownsMat: Boolean = true
 ) : IPattern {
+    private data class MatWithAlpha(val mat: Mat, val alpha: Mat?)
+
     private companion object {
-        fun makeMat(stream: InputStream): Mat {
+        fun makeMat(stream: InputStream): MatWithAlpha {
             val byteArray = stream.readBytes()
+            MatOfByte(*byteArray).use {
+                val grayScale = Imgcodecs.imdecode(it, Imgcodecs.IMREAD_GRAYSCALE)
```
MathewSachin, line 30: Add a comment here that imread grayscale and converting to grayscale from imread unchanged is different, or someone might do that again.

Other than that, tested that works on both NA and JP on wide screen.

MathewSachin (approved):

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
MathewSachin: Also, make sure to mention alpha matching in the commit message.

app/src/main/java/com/mathewsachin/fategrandautomata/imaging/DroidCvPattern.kt
```diff
@@ -27,24 +27,28 @@ class DroidCvPattern(
...
+                                Mat().apply { Core.extractChannel(decoded, this, 3) }
+                            val minMax = Core.minMaxLoc(alphaChannel)
+                            if (minMax.minVal.equals(255.0)) {
+                                //every pixel has 0 transparency, alpha is useless
+                                alphaChannel.release()
+                                alphaChannel = null
+                            }
+                            Imgproc.cvtColor(decoded, grayScale, Imgproc.COLOR_BGRA2GRAY)
                         }
+                        3 -> Imgproc.cvtColor(decoded, grayScale, Imgproc.COLOR_BGRA2GRAY)
```
MathewSachin, line 47: RGBA would be for 4 channels. RGB for 3.

reconman: Don't you mean BGR? According to the internet, OpenCV reads images in BGR format: https://stackoverflow.com/a/59851524/5971497

That's not the only place where I've found this info.

PR #821, opened 2021-08-11
Issue Template

the3dsandwich:
Too many people don't know how to properly check checkboxes. I was thinking if we can make the template more self-explanatory so people can have an easier time creating issues (and we might have less dumb issues).

Mainly I want to add a example check so people can follow, and some more explanatory text saying which section can be deleted and which ones should stay.


MathewSachin:
I just saw this: [Creating issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms).
This would be so awesome once it gets out of Beta.

reconman:
I would personally just use Github Forms, even if it's in beta

the3dsandwich:
Whoa that seem nice! I'll take a look tomorrow and see if I can come up with something. Thanks!

reconman:
See #823 for a Github forms solution

Issue #772, opened 2021-07-08
Still not executing the correct script after changing Stage Counter Similarity


advancedorange:
Before opening an issue:
- [x] I tested the latest release
- [x] I looked at other issues (even the closed ones)
- [x] I read the [Troubleshooting Guide](https://github.com/MathewSachin/Fate-Grand-Automata/wiki/Troubleshooting)
- [x] I'll fill this template properly

> Your issue will get closed/locked by the bot in case proper info isn't provided
If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

**Describe the bug**
The old report got closed, but the issue started reappearing

Previous description:
A few days ago I switched to a new node to farm and created a new script, but this script does not what its supposed to do.
Most of the time the script for wave 3 gets activated on wave 2, sometimes the script even executes the steps for wave 2 on the second enemy on the first wave, followed by the steps for wave 3 on the third one (still wave 1). One time this was happening I think I noticed FGA trying to tap the skill for wave 2 right after it pressed the attack button for wave 1 (before the cards even popped up).

Setting the Stage Counter Similarity to 75% did not fully fix this problem
Today I once again had a run during which everything was executed on the first wave

Script: > i h g1 d ,#, f 5 ,#, g1 i h k1 c 4 ,#, 
Party: Summer Scathach, Yu Meiren, Waver 
Node: Fallen Babylon
Server: NA
Link to reddit: [Here](https://old.reddit.com/r/grandorder/comments/nyvt9g/help_and_question_thread_june_13_2021/h29ir3m/ )

**Screenshots**
x

**Video**
x

**Device Info (please fill at-least 'Device model'):**
 - Device model: [Honor 8]
 - Screen size: [1920x1080]
 - Android Version: [Android 8 (?)]
 - RAM: [4GB]

reconman:
> One time this was happening I think I noticed FGA trying to tap the skill for wave 2 right after it pressed the attack button for wave 1 (before the cards even popped up).

Sounds like you need to increase `Wait after click`.

MathewSachin:
> **Screenshots**
If applicable, add screenshots to help explain your problem.
It is recommended that you at-least add one screenshot of FGO running on your device.
>
>**Video**
If relevant, include a video of the script running on your device with `Debug mode` ON
Recording functionality is available within the app in `More options`.

☝️

advancedorange:
How much should I increase Wait after click?

I'll try to make a video tomorrow

[...]

PR #823, opened 2021-08-11
Use Github forms for templates

reconman:
You can check out my forms in action over at https://github.com/reconman/Fate-Grand-Automata/issues/new/choose

You can also see them when you browse the files, although it will show a table at the top with some template properties: https://github.com/Fate-Grand-Automata/FGA/blob/issue-forms/.github/ISSUE_TEMPLATE/bug_report.yml


MathewSachin:
1. Screen size and RAM can be made optional. Not everyone would know screen size atleast.
2. Ask for the game server in a dropdown with possible values.
3. Should we ask the Canary version number while also telling for latest release? There were some cases like #772 who claimed to be on latest release, but were not even on Canary and using months old build from PlayStore.

reconman:
I addressed your comments, you can take a look again.

MathewSachin (approved):

PR #929, opened 2021-10-30
Refresh screen only after the right class is selected

### LINKED ISSUE #924: Make FGA check the selected class when selecting supports even if there were no supports found after the filter
### Preparation

### Describe the feature

This might fit better as a bug, but I'm not sure how intended it is, so I decided to write it as a suggestion. Currently, if the support list contains no servants (because the event filters filter out everyone on the currently selected class's list) the script immediately goes to refresh the support list, and repeat that until there's at least one result.

The problem is when the currently selected class list is not the one that the script is supposed to check. For example, let's say the script is set to get the first result of the caster support list, but the node is actually a lancer node, so the game defaults to displaying the saber support list at first. If the saber support list has no supports appearing, the script will keep refreshing the list until it gets a servant on the saber support list, as opposed switching to the caster support list, and _then_ checking if there are supports available or if it has to refresh.

reconman:
Fixes #924

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.2'
+        classpath 'com.android.tools.build:gradle:7.0.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "com.google.dagger:hilt-android-gradle-plugin:$hilt_version"
         // NOTE: Do not place your application dependencies here; they belong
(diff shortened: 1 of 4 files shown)


reconman:
I've tested without All mode and it worked fine. I'm just not really happy with copy pasting the code to First and Specific support. But I can't put it the SupportSelectionProvider interface as a default method because `images` and `locations` don't exist there. I also can't put it into SupportRefresher, since only Support has a reference to that object.

MathewSachin:
I had also made changes for this 😅 also taking care of 'also check all'. need a bit more time to test. will put up a review tomorrow.

>  I'm just not really happy with copy pasting the code to First and Specific support. But I can't put it the SupportSelectionProvider interface

If we really needed to do this, we could add a wrapper class, but, I went ahead and changed the wait loop to support 'also check all'.

MathewSachin:
Closing in favour of #933 😅

PR #951, opened 2021-11-16
New UI support for EN server

### LINKED ISSUE #948: Widescreen support for NA
### Preparation

### Describe the feature

https://fate-go.us/news/?category=UPDATES&article=%2Fiframe%2F2021%2F1114_gameupdate_notice%2F

NA rolled out the widescreen changes today

Kobata:
Should fix #948

Also a probably less prone to errors support.png (after a few tries, I just copied the one from JP -- the font's the same anyway)

The location change is because it kept trying to hit the drop list starting on the main menu, some others might need to be checked more.


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/locations/Locations.kt
```diff
@@ -31,7 +31,7 @@ class Locations @Inject constructor(
 
     val menuSelectQuestClick =
         (if (isWide)
-            Location(-410, 440)
+            Location(-600, 440)
```
MathewSachin, line 34: Hey, thanks for the PR.
I just have a concern about changing this value by too much.
We had some issues in past related to this which I'm not able to find right now.
There was some raid event which had a delay in opening AP refill dialog, we had to shift the click to be between the AP dialog and the node info button to solve that.
See if you can get by with a little less change in the value.

potchy:
Is Tawain now the only server still using the old UI?
I'm seeing a opportunity to get rid of that `isNewUI` flag pretty soon. 👀

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/locations/Locations.kt
Kobata: Did some quick testing with other values, `-460` seems to be fairly close to the minimum, and is outside the AP box on my device at least.

MathewSachin (approved):

MathewSachin:
> Is Tawain now the only server still using the old UI?

yup, @the3dsandwich, how much behind is the TW server from JP?

> I'm seeing an opportunity to get rid of that isNewUI flag pretty soon. 👀

Rest of the code would still stay same even then. `isWide` checks would always be needed. 😢

Santicx:
can confirm it still does not work on all screen.
142001702-e128caba-165d-4fb9-993b-f06ecacdf1d0
142001689-1a837c32-8b65-47cd-bc62-b1c68a4038b6
as you can see my "widescreen" is not fully covered which I will assume it bugs the AI.
the phone is Asus Rog 5

Santicx:
I can confirm that  the AI is still not recognizing  the screen
Screenshot_20211116-182250421
Screenshot_20211116-182210910

the "widescreen fix" does not cover the full screen and I guess thats

 Phone is Asus Rog phone 5

potchy:
Have you seen the troubleshoot section of the wiki, [CASE 2: Your phone has borders around FGO even at the bottom of screen](https://github.com/Fate-Grand-Automata/FGA/wiki/Game-Area-detection#case-2-your-phone-has-borders-around-fgo-even-at-the-bottom-of-screen), specifically?

Santicx:
> Have you read the troubleshooting section of the wiki, [CASE 2: Your phone has borders around FGO even at the bottom of screen](https://github.com/Fate-Grand-Automata/FGA/wiki/Game-Area-detection#case-2-your-phone-has-borders-around-fgo-even-at-the-bottom-of-screen), specifically? If your phone is rooted, you can install the "Fullscreen/Immersive Gestures (Q-S)" Magisk module to get rid of the bottom border while still having gesture navigation enabled.


I tried what you said 1st my phone is not rooted so I can not get rid of the bottom border
but I changed the Gestures to 3-Buttons and the game stucks at this stage of the game. Its kinda disappointing this porblem occuring 
Screenshot_20211116-194520502

potchy:
Please continue with the discussion in #955. It seems to be the same problem you're having.

the3dsandwich:
@MathewSachin Wait we're the only ones without the wide screen update? 

We're approximately 1 and a half years behind JP but as TW gets the source app from CN we should get it in a few months.

PR #1025, opened 2021-12-19
Fix "Failed to Determine" for buffed servants

### LINKED ISSUE #920: Failed to determine card type on super-buffed Lancelot on T3
### Preparation

- [X] I tested the [latest release](https://github.com/Fate-Grand-Automata/FGA/releases/tag/canary)
- [X] I looked at other issues (even the [closed ones](https://github.com/Fate-Grand-Automata/FGA/issues?q=is%3Aissue+is%3Aclosed))
- [X] I read the [Troubleshooting Guide](https://github.com/MathewSachin/Fate-Grand-Automata/wiki/Troubleshooting)

### FGO server

NA

### FGA build number

1537

### Describe the bug

On turn 3, I'm getting "failed to determine card type" error on Lancelot. It works every other turn, but it always fails on turn 3. I've added a video of what happens in turn 3 when Lancelot face cards are there.

It's happened on every Saber Wars 2 nodes I've done since I updated last night. It also happens on non-event nodes, so it has nothing to do with the bonus damage buffs Lancelot has plus the CE.

Buff order for Lancelot goes
**T1:** Paracelsus Arts, Waver DEF, Waver ATK
**T2:** Paracelsus Guts, Paracelsus NP Gain, Waver Crit, Skadi Quick, Skadi Quick Crit, Lancelot NP Gain (_FALLS OFF START OF T3_), Crit Strength, Crit Star Drop, Crit Star Gather
**T3:** Command Code ATK buff

I saw similar issues with buffs messing with card matching, but I noticed they were said to be fixed. I don't know how fringe this case is, but it's still happening seemingly. Every setting is default as far as I remember.

Team looks as such:
Screenshot_2021 10 26_07 39 05 987

### Video

https://streamable.com/246c5d

### Device model

Bluestacks 5 Pie

### Android version

9

### Screen size

_No response_

### RAM

_No response_

reconman:
Fixes #920 
Adjusts the message for no EXP cards on screen.

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/util/KnownException.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/util/KnownException.kt
@@ -2,24 +2,27 @@ package com.mathewsachin.fategrandautomata.util
 
 import com.mathewsachin.fategrandautomata.SupportImageKind
 
-class KnownException(val reason: Reason): Exception(reason.msg) {
+class KnownException(val reason: Reason) : Exception(reason.msg) {
     sealed class Reason(val msg: String) {
-        class CouldNotCrateDirectory(name: String): Reason("Couldn't create directory: '$name'")
-        class CouldNotCrateFile(name: String): Reason("Couldn't create file: '$name'")
-        object CouldNotOpenFileForRecording: Reason("Couldn't open file for recording")
+        class CouldNotCrateDirectory(name: String) : Reason("Couldn't create directory: '$name'")
+        class CouldNotCreateFile(name: String) : Reason("Couldn't create file: '$name'")
+        object CouldNotOpenFileForRecording : Reason("Couldn't open file for recording")
         class CouldNotOpenSupportFileForReading(
             kind: SupportImageKind,
             name: String
-        ): Reason("Couldn't open file for reading: [$kind] '$name'")
+        ) : Reason("Couldn't open file for reading: [$kind] '$name'")
+
         class CouldNotOpenSupportFileForWriting(
             kind: SupportImageKind,
             name: String
-        ): Reason("Couldn't open file for writing: [$kind] '$name'")
-        object CouldNotCreateDropScreenshotFile: Reason("Failed to create drop screenshot file")
+        ) : Reason("Couldn't open file for writing: [$kind] '$name'")
+
+        object CouldNotCreateDropScreenshotFile : Reason("Failed to create drop screenshot file")
         class SupportFolderIsEmpty(
             kind: SupportImageKind,
             name: String
-        ): Reason("[$kind] folder: '$name' is empty!")
-        class FailedRootPermission(e: Exception): Reason("Failed to get Root permission: ${e.message}")
+        ) : Reason("[$kind] folder: '$name' is empty!")
+
+        class FailedRootPermission(e: Exception) : Reason("Failed to get Root permission: ${e.message}")
     }
 }
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/util/KnownException.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/util/KnownException.kt
@@ -4,7 +4,7 @@ import com.mathewsachin.fategrandautomata.SupportImageKind
 
 class KnownException(val reason: Reason) : Exception(reason.msg) {
     sealed class Reason(val msg: String) {
-        class CouldNotCrateDirectory(name: String) : Reason("Couldn't create directory: '$name'")
+        class CouldNotCreateDirectory(name: String) : Reason("Couldn't create directory: '$name'")
         class CouldNotCreateFile(name: String) : Reason("Couldn't create file: '$name'")
         object CouldNotOpenFileForRecording : Reason("Couldn't open file for recording")
         class CouldNotOpenSupportFileForReading(
--- a/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoGiftBox.kt
+++ b/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoGiftBox.kt
@@ -18,7 +18,12 @@ class AutoGiftBox @Inject constructor(
     api: IFgoAutomataApi,
     private val connectionRetry: ConnectionRetry
 ) : EntryPoint(exitManager), IFgoAutomataApi by api {
-    class ExitException(val pickedStacks: Int) : Exception()
+    sealed class ExitReason {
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/mathewsachin/fategrandautomata/util/KnownException.kt
```diff
@@ -2,24 +2,27 @@ package com.mathewsachin.fategrandautomata.util
 
 import com.mathewsachin.fategrandautomata.SupportImageKind
 
-class KnownException(val reason: Reason): Exception(reason.msg) {
+class KnownException(val reason: Reason) : Exception(reason.msg) {
     sealed class Reason(val msg: String) {
-        class CouldNotCrateDirectory(name: String): Reason("Couldn't create directory: '$name'")
-        class CouldNotCrateFile(name: String): Reason("Couldn't create file: '$name'")
-        object CouldNotOpenFileForRecording: Reason("Couldn't open file for recording")
+        class CouldNotCrateDirectory(name: String) : Reason("Couldn't create directory: '$name'")
```
MathewSachin, line 7: Please correct this one too

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoGiftBox.kt
```diff
@@ -30,7 +30,7 @@ class AutoGiftBox @Inject constructor(
     override fun script(): Nothing {
         val xpOffsetX = (locations.scriptArea.find(images[Images.GoldXP]) ?: locations.scriptArea.find(images[Images.SilverXP]))
             ?.region?.center?.x
-            ?: throw Exception("Couldn't find Embers on screen. This shouldn't happen.")
+            ?: throw Exception("Couldn't find Embers on screen. Make sure that you've filtered for EXP cards.")
```
MathewSachin, line 33: Can you add this to KnownException.
That way the stack trace won't be shown.

reconman: KnownException is part of `app`, this code is part of `scripts`.

MathewSachin:
I'm not very confident about just shifting the regions down. It might interfere with the NP name.
I think we might have to reduce the height or the similarity.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoGiftBox.kt
MathewSachin: Oh right, may bad.
Do we have a ExitReason thing for GiftBox like we have for Battle and other stuff?

reconman: We only have ExitException with the number of selected cards.

MathewSachin: Hmm, we don't have it yet.
If you don't mind, can you add that.

This used to be an unexpected exception until the option to receive embers was added to lottery script.

reconman:
> I'm not very confident about just shifting the regions down. It might interfere with the NP name.

Blue is the old location, the other rectangle is the new area. It's nowhere near the NP name:
screenshot

MathewSachin:
OMG! I don't even remember keeping that region high up. Maybe because some servants' necks were not wide enough 🤣

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/ServantTracker.kt
```diff
@@ -181,53 +180,51 @@ class ServantTracker @Inject constructor(
...
+            val matchedServants = ownedServants.mapValues { (_, image) ->
+                locations.attack.servantMatchRegion(card)
+                    .find(image, 0.5)?.score ?: 0.0
+            }.filterValues { it > 0.0 }
+
+            matchedServants
+                .maxByOrNull { it.value }
+                ?.let { (teamSlot, _) ->
+                    // add the card to the matched Servant
+                    result.getOrPut(teamSlot, { mutableSetOf() }) += card
```
MathewSachin, line 216: You can use the groupBy operator instead of the
 loop.
Also, see if you can avoid MutableSet and just use Set.

reconman:
The most recent commit was because of this problem:
screenshot

Resist isn't detected correctly if you have CCs on your cards. I made the rectangle a bit higher.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/modules/ServantTracker.kt
```diff
@@ -186,36 +186,34 @@ class ServantTracker @Inject constructor(
...
-
                 result[supportSlot] = matched
             }
         }
 
         val ownedServants = faceCardImages
             .filterKeys { it != supportSlot && it in deployed.values }
-        for (card in cardsRemaining) {
+        cardsRemaining.groupBy { card ->
             // find the best matching Servant which isn't the support
```
MathewSachin, line 205: It would be easier to read if groupBy started from next line, the operators would be indented on the same level then. Same for mapValues down below.
But this is just me nitpicking

MathewSachin:
```kt
cardsRemaining
    .groupBy { card ->
        // find the best matching Servant which isn't the support
        ownedServants
            .mapValues { (_, image) ->
                locations.attack.servantMatchRegion(card)
                    .find(image, 0.5)?.score ?: 0.0
            }
            .filterValues { it > 0.0 }
            .maxByOrNull { it.value }
            ?.key
    }
    .filterKeys { it != null }
    .entries
    .associateTo(result) { (key, value) -> key!! to value.toSet() }
```

reconman: Done, didn't know what caused the extra indents at the bottom.

MathewSachin: The IDE only indents correctly if the operators are on new line for some reason.

Can you do the same for mapValues too like in my previous comments? It looks weird right now 😅

reconman: Done

MathewSachin (approved):

PR #1046, opened 2021-12-25
Let MediaProjection do the image scaling

MathewSachin:
It's a minor performance improvement to let MediaProjection write to a 720p surface instead of resizing via OpenCV.


MathewSachin:
@reconman I've tested for MediaProjection, can you help test if the root mode still works after this change? Thanks.

reconman:
Works perfectly.

PR #1137, opened 2022-03-27
Create error toasters in the UI thread

reconman:


app/src/main/java/com/mathewsachin/fategrandautomata/runner/ScriptManager.kt
```diff
@@ -108,8 +109,10 @@ class ScriptManager @Inject constructor(
                         }
                     } catch (e: Exception) {
                         val msg = "Failed to stop recording"
-                        Toast.makeText(service, msg, Toast.LENGTH_SHORT).show()
                         Timber.e(e, msg)
+                        (context as Activity).runOnUiThread {
```
MathewSachin, line 113: ApplicationContext isn't going to be an Activity.
Use `withContext(Dispatchers.Main)` instead.

MathewSachin (approved):

Issue #1212, opened 2022-06-14
TW Fullscreen


the3dsandwich:
### Preparation

- [X] No one has asked for this feature before in other issues (even the closed ones)

### Describe the feature

TW finally received its fullscreen update. I’ll make a PR later if nobody’s on it yet.

PR #1213, opened 2022-06-14
Update for TW (Inc. New UI and Localization)

the3dsandwich:
Mentioned in #1212. Now that TW is finally up to date with the rest of the servers, we can update the settings for FGO TW.

BTW may I also have an update on what's currently being added to FGA so I can help with images or translation?

Thanks!


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/prefs/IPreferences.kt
```diff
@@ -52,4 +52,4 @@ interface IPreferences {
 
 val IPreferences.wantsMediaProjectionToken get() = !useRootForScreenshots
 
-val IPreferences.isNewUI get() = gameServer in listOf(GameServerEnum.Jp, GameServerEnum.En, GameServerEnum.Kr, GameServerEnum.Cn)
\ No newline at end of file
+val IPreferences.isNewUI get() = gameServer in listOf(GameServerEnum.Jp, GameServerEnum.En, GameServerEnum.Kr, GameServerEnum.Cn, GameServerEnum.Tw)
```
reconman: isNewUi is no longer needed since that condition now always evaluates to `true`. All code lines for `!isNewUi` can be removed now.

reconman: I've committed the removal of that code. One thing I noticed is that we're not sure if CN displays FGO over notches or not. See CutoutManager line 77.

Does TW need that?

the3dsandwich: > I've committed the removal of that code. One thing I noticed is that we're not sure if CN displays FGO over notches or not. See CutoutManager line 77.
> 
> Does TW need that?

I cannot be sure about notches, since I don't have a phone with a notch to test atm. Is there any other TW player with a notched phone to confirm?

lycsjm: FGO TW display black bolder in notch area on my old phone.

the3dsandwich:
Screenshot_20220615-234032

I have a screenshot. Is this the one?

I was originally thinking about pushing some localized text before asking to merge PR, but since S**tfest will start tomorrow I think it might be better that we prioritize TW's screen size problem above texts.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/prefs/IPreferences.kt
reconman: I've changed the code to be the same for CN and TW. Assume that the notch area is not used by FGO.

reconman (approved):

PR #1243, opened 2022-07-10
Localization: zh_CN

MetLee:
Fixed some typos(also in zh_TW) and added some localized texts in zh_CN.

Some features I never used were not tanslated: `p_roll_limit` and `overlay_notification_text`.

https://github.com/Fate-Grand-Automata/FGA/blob/35279ab7dbbbae2c91bc5f2b4c80a3ebff419c7e/app/src/main/java/com/mathewsachin/fategrandautomata/ui/exit/BattleExit.kt#L62-L63

and

https://github.com/Fate-Grand-Automata/FGA/blob/35279ab7dbbbae2c91bc5f2b4c80a3ebff419c7e/app/src/main/java/com/mathewsachin/fategrandautomata/ui/launcher/LotteryLauncher.kt#L70

I am not familiar with Kotlin, could you add those phrases into `localized.xml` ?


app/src/main/res/values-b+zh+TW/localized.xml
```diff
@@ -84,7 +84,7 @@
     <string name="p_battle_config_summary">設定周回腳本</string>
     <string name="p_more_options">更多選項</string>
 
-    <string name="p_runs">上限</string>
+    <string name="p_limit">上限</string>
```
reconman, line 87: p_limit was already translated, but you removed the translation for p_runs.

app/src/main/res/values/localized.xml
```diff
@@ -95,6 +95,7 @@
     <string name="p_runs">Runs</string>
     <string name="p_roll_limit">Limit number of rolls</string>
     <string name="p_prevent_lottery_box_reset">Prevent box reset</string>
+    <string name="p_receive_embers">Receive embers</string>
```
reconman, line 98: Please translate p_receive_embers, stop_after_this_run, tapper_service_description, overlay_notification_text and p_roll_limit.

reconman (changes requested):

app/src/main/res/values-b+zh+TW/localized.xml
MetLee: Yes I found the other one after receiving your comment. 
`p_runs` -> `上限` was a wrong translatioin existing in both `zh_CN` and `zh_TW` (maybe those were someone's typos), I (as a mandarin-speaker) corrected the translation in `zh_CN`, but I don't know the appropriate translation in `zh_TW` (though they look similar). 
Users will be confused about a wrong translation. Could you removed that line?

app/src/main/res/values/localized.xml
MetLee: Done

MetLee: Make some words I used in last commit more formal and fixed a typo (again).

MetLee:
I found some hard-coding sentence:

https://github.com/Fate-Grand-Automata/FGA/blob/61cf9636cdd591c090850e3a1cf31db07bf60038/app/src/main/java/com/mathewsachin/fategrandautomata/ui/more/BattleGroup.kt#L101-L102

https://github.com/Fate-Grand-Automata/FGA/blob/cd8f96085f0c5c81a1b3dc772f47bfb313592def/app/src/main/java/com/mathewsachin/fategrandautomata/ui/more/AdvancedGroup.kt#L81

https://github.com/Fate-Grand-Automata/FGA/blob/cd8f96085f0c5c81a1b3dc772f47bfb313592def/prefs/src/main/java/com/mathewsachin/fategrandautomata/prefs/core/GameAreaMode.kt#L4-L7

https://github.com/Fate-Grand-Automata/FGA/blob/cd8f96085f0c5c81a1b3dc772f47bfb313592def/app/src/main/java/com/mathewsachin/fategrandautomata/ui/more/AdvancedGroup.kt#L108

app/src/main/res/values-b+zh+TW/localized.xml
reconman: @the3dsandwich What's your opinion on this?

app/src/main/res/values/localized.xml
reconman: tapper_service_description is still missing.

reconman:
I've added more translatable strings here https://github.com/Fate-Grand-Automata/FGA/commit/ea8ebf4305cbddc6761e3450e8d3d8ac233daf70#diff-2d207ac16b141c7910606683e804e844a24ef211c6210e4f1842673f396495e8

app/src/main/res/values-b+zh+TW/localized.xml
the3dsandwich: I looked into `BattleExit.kt` and found that it's used as `"${stringResource(R.string.p_runs)}: $runs"`. In this regard I think both CN and TW can change to `已執行次數: $runs` (and its Simplified Chinese variant)

reconman:
Revert your last commit, the quotes are needed to keep the space at the end.

MetLee:
And what's the "NP-matching card shuffle" mean?

reconman:
There's only one string without translation: p_thresholded_stage_counter.

This is the wikipedia article on thresholding, since it's hard to explain: https://zh.wikipedia.org/wiki/%E4%BA%8C%E5%80%BC%E5%8C%96

reconman:
@MetLee Reminder

MetLee:
> @MetLee Reminder

I am not good at image processing. I guess that when the switch on, FGA will use the binary image (not the RGB image or the grayscale image) to detect the stage counter, for reducing consumption and rapid response. Is it right?

reconman:
I linked the Wiki page because you can reuse words from there.

The full explanation is:
FGA normally takes screenshots of the stage counters and compares the images with each other. If they're similar, we're on the same stage. Otherwise, the stage changed.

During this comparison, the images are first turned to shades of gray. Then a complicated algorithm is used to calculate the similarity by comparing each pixel and calculating the difference.

The "thresholded stage counter" works a bit different: The old image is still turned gray, but the new one is turned into the whitest white and blackest black. Anything lighter than a certain gray tone is turned white, the rest is turned black. Then the usual algorithm is used to compare the old gray image with the new black and white image.

I can't tell you if it works better than the old method or not.

MetLee:
OK, I understood. So it is just a feature which we don't know how well it works. This function only affects the new image and the old one is still gray scale.

MetLee:
I think that the user don't have to know how the feature works (but translator should figure it out), so I focus the translation on the binary image.

reconman (approved):

PR #1299, opened 2022-08-19
Dynamic skill button sizes in Command Editor

### LINKED ISSUE #1290: Improve Battle command UI
### Preparation

### Describe the feature

The space between buttons and the size of the button should not be static because the screen size of the phones are different. It should depend on the height and width of the screen because some devices have small screens, it would cause some buttons to disappear and that is not good for user experience.
The screenshot is from my device which is samsung galaxy j7 prime.
Screenshot_20220814-205232_FGA_CI
.

reconman:
Fixes #1290


vietng322611:
You should change skill button, master skill and button with icon size too

reconman:
> You should change skill button, master skill and button with icon size too

I have no idea how to do that. I tried for 1 hour yesterday.

PR #1304, opened 2022-08-22
Add Korean translation about new added materials

item4:


item4:
comment: there is no official translated text for `Twilight Ceremonial Blade` and `Unforgettable Ashes` yet.

PR #1337, opened 2022-08-27
Add inventory full check function for fix logic

item4:
By this PR, FP Gacha scripts changes::

1. Can start already run cost-free FP, click position is move right little for cover free button and 2000 FP button both.
2. Can check inventory is full, and if inventory was full, exit script.
3. Use one loop logic instead one if and one while. It make easy to manage code.
4. Replace KR server `inven_full.png` file to grayscale.


item4:
I think I choose wrong solution but it did not work fine. old version can not detect inventory full.

reconman:
To avoid spam from commit pushes, could you please build and test the app locally with Android Studio? Right now, it doesn't even compile.

You only need to change one line in app/build.gradle:
```
versionCode Integer.valueOf(System.env.FGA_VERSION_CODE ?: 1)
```

Replace the `1` at the end with the version currently installed on your device.

item4:
Yes I will

item4:
@reconman Okay, I setup Android Studio on my Windows PC and build & test locally. But I cannot test Jp and En server. I updated description of this PR. please check is it right

item4:
@reconman review please 🙏

item4:
Close PR, I will remake that

PR #1376, opened 2022-09-09
Fix inventory full checker

item4:
I was make PR changing FP gacha, but I feel something strange and I make new PR.

1. I don't know reason why is working wrong, but I must change `inv_full.png` for Kr server because FGA can not detect `Reinforce` button.
2. There is no requirement for changing detection area, only need setting similarity for detecting. So I did not change location definition at this PR.
3. The logic about inventory full checking is also used for Auto battle. Old PR cover FP gacha only. I fix that for changing also used for AB.
4. I exclude some changes about FP gacha. Because:
    1. There is some difference among each servers about first free FP gacha screen. Kr service has FP cost icon on screen but others is not.
    2. I need fix inventory full checking much more then FP gacha. So I drop or split them. If you want these contents also, I will make PR after merging this.

This PR tested on Kr server and it's fine.


scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoBattle.kt
```diff
@@ -28,7 +28,8 @@ fun IFgoAutomataApi.isInSupport(): Boolean {
 fun IFgoAutomataApi.isInventoryFull() =
     // We only have images for JP, NA and KR
     prefs.gameServer in listOf(GameServerEnum.En, GameServerEnum.Jp, GameServerEnum.Kr)
-            && images[Images.InventoryFull] in locations.inventoryFullRegion
+            && locations.inventoryFullEnhancementButtonRegion.exists(images[Images.InventoryFull], similarity = 0.23)
```
reconman: 23 % similarity is incredibly low. I think you had to set it that low because you used `&&` instead of `||` in line 32.

Can you try again with the changes from my commit?

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/locations/Locations.kt
```diff
@@ -21,7 +21,8 @@ class Locations @Inject constructor(
     val continueRegion = Region(120, 1000, 800, 200).xFromCenter()
     val continueBoostClick = Location(-20, 1120).xFromCenter()
 
-    val inventoryFullRegion = Region(-280, 860, 560, 190).xFromCenter()
+    val inventoryFullEnhancementButtonRegion = Region(-280, 860, 560, 190).xFromCenter()
+    val inventoryFullSellButtonRegion = Region(-880, 860, 560, 190).xFromCenter()
```
reconman: If I understand correctly, you had to add this Region because the button is located somewhere else for FP summons?

Because all values are the same except for X, I combined both regions into one big one by extending the old one 600 to the left.

item4:
@reconman It was merged on my branch and pr already by your changes. I will test it and attach video.

item4:
## Running on [2ce97bf](https://github.com/Fate-Grand-Automata/FGA/pull/1376/commits/2ce97bf7e2ee23daecdf9f159b31419f675c8343)

https://user-images.githubusercontent.com/559952/189509258-a8c2a925-5412-4651-b2b0-ed8ff265f957.mp4

item4:
## What I want to detect for catch inventory was full

image

## But

Can not catch it. So I decrease similarity. It was work, but has side-effects

## Expected screens, but can not dectect now





## Wrong detection on low similarity



## What really I want to detect

### Case 1. detect text correctly

image
image

### Case 2. detect two buttons

image
image

## Problems what I can not solve

1. Without changing region size or similarity , I can not detect inven_full.png from detecting region.
2. I tried changing region size to fit to kr text but it was not working. (I don't know reason why)
3. I tried to decrease similarity but it cause false-positive.
4. I tried to replace inven_full.png to more lager size for fitting without change region size. but that also not working

reconman:
I think you didn't notice that your image is much bigger than the button in the screenshot.

This is the inven_full.png next to the actual button.

You probably forgot to take the image in 720p.

reconman:
Please test the latest build again.

item4:
@reconman Wow. Good It works! I did not know pattern image must capture in 720p. but how can i take 720p correctly? resize after screenshot? or other method here?

reconman:
Yes, you need a photo editor like Adobe Photoshop or the free online Photopea to scale your image down.

Issue #998, opened 2021-12-10
Adjust CE Bomb for wide screen, add documentation


reconman:
### Preparation

- [X] No one has asked for this feature before in other issues (even the closed ones)

### Describe the feature

The current CE bomb feature is not documented, afaik. Also, the X offsets don't work for fullscreen.

To Dos:
- [ ] Document the feature
- [ ] Add `isWide` checks for the 3 locations. We also need to check if the wide screen X offset is the same for NA and JP.
- [ ] Adjust the script for loading times, since any loading can cause the script to break.

@TheIanator Any chance you can help out?

For the documentation, I'd need a description in Markdown format. There are a few markdown editors available on the internet, including the one for writing Github comments.

KaitoKid:
Maybe I'm on the wrong release/branch but I don't actually see the CE bomb feature. Happy to work on a draft of documentation that addresses risk (or minimizes risk as much as possible) and usability

reconman:
https://github.com/Fate-Grand-Automata/FGA/blob/master/scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/entrypoints/AutoCEBomb.kt#L64

Willektron:
I asked about this in the Gamepress thread a couple months ago; there's a bit of information in the comments at the beginning of the file, and a good bit more in the original issue (by TheIanator, who I recognize you've already pinged):

https://github.com/Fate-Grand-Automata/FGA/issues/733

What would documentation for this entail? Adding a page with info to the wiki?

reconman:
> What would documentation for this entail? Adding a page with info to the wiki?

Exactly, but only FGA members can edit wiki pages.

That's I will have to copy the guide.

Some screenshots in the guide would be helpful, I think.

Ansem93:
Holy! I just discovered this feature!

Nameless-Blob:
I can't use the feature, i have a wide screen phone, is there any way for me to fix the script to launch this utility?

gary-stu:
Hey there.

Since it was broken on wide phones and I wanted to use a CE bomb script, I just rewrote something for my personal use. 
I threw something up in four hours, then spent a couple more testing on several phones and doing a couple adjustments .
It did work as intended on all three wide phone I tested, as well as an old non wide phone.

The script itself does the following: 
- Tries to select a CE to enhance, going in order from top left CE to bottom right CE
[...]

PR #1383, opened 2022-09-17
CE bomb improvements by gary-stu

ArthurKun21:
the patch was here https://github.com/Fate-Grand-Automata/FGA/issues/998#issuecomment-1248435280 by gary-stu

PR #1385, opened 2022-09-17
CE bomb improvements by gary-stu

matthewasage:
Patch submitted in #998 by gary-stu, also added fix for inconsistent OK button detection (also by gary-stu).


reconman:
Closing in favor of #1383

PR #1490, opened 2022-12-18
Adds CE drops to between run toast

patwakeem:
- Adds CE drop text to between run toast. 
- Helpful for when a user doesn't want to pause scripts to check if a CE has dropped.

### DIFF
--- a/app/src/main/java/com/mathewsachin/fategrandautomata/util/ScriptMessages.kt
+++ b/app/src/main/java/com/mathewsachin/fategrandautomata/util/ScriptMessages.kt
@@ -119,7 +119,8 @@ class ScriptMessages @Inject constructor(
             is ScriptNotify.BetweenRuns -> {
                 val msg = makeRefillAndRunsMessage(
                     timesRefilled = action.refills,
-                    timesRan = action.runs
+                    timesRan = action.runs,
+                    ceDrops = action.ceDrops
                 )
 
                 toast(msg)
@@ -128,7 +129,8 @@ class ScriptMessages @Inject constructor(
 
     private fun makeRefillAndRunsMessage(
         timesRan: Int,
-        timesRefilled: Int
+        timesRefilled: Int,
+        ceDrops: Int
     ) = buildString {
         val refill = prefs.refill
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/mathewsachin/fategrandautomata/util/ScriptMessages.kt
```diff
@@ -142,6 +144,12 @@ class ScriptMessages @Inject constructor(
             )
         }
 
+        if (ceDropCount > 0) {
+            appendLine(
+                context.getString(R.string.ces_dropped, ceDropCount)
```
reconman, line 149: ces_dropped needs to be added to localized.xml

reconman (changes requested):

app/src/main/java/com/mathewsachin/fategrandautomata/util/ScriptMessages.kt
patwakeem: Oof ill take a look, thanks!

reconman: The main one at https://github.com/Fate-Grand-Automata/FGA/blob/master/app/src/main/res/values/localized.xml is the most important one.

Without that, the app won't compile.

app/src/main/res/values-b+zh+CN/localized.xml
```diff
@@ -288,6 +288,8 @@
 
     <string name="times_rolled">已抽取 %d 次</string>
 
+    <string name="ces_dropped">礼装掉落: %d</string>
```
reconman, line 291: Are you sure that this is correct? If not, leave it to the Chinese devs who occasionally translate new features.

patwakeem: I'll roll this one back, sorry android studio was giving me an error saying it wasnt translated to chinese.

app/src/main/java/com/mathewsachin/fategrandautomata/util/ScriptMessages.kt
patwakeem: Small question, I reused this message:

https://github.com/Fate-Grand-Automata/FGA/blob/master/app/src/main/res/values/localized.xml#L292

That should cover the message or should I add another one for this toast specifically? I'm willing to do it either way.

reconman: Ah sorry, didn't notice that you reused an old one.

reconman (approved):

Issue #1532, opened 2023-02-08
New Skill Confirmation for Kukulkan


imniwa:
### Preparation

- [X] No one has asked for this feature before in other issues (even the closed ones)

### Describe the feature

Request new confirmation skill like space ishtar. this servant have 2 option ltr (cancel, use without absorb star, use absorb 10 star). 

![Screenshot_2023-02-08-07-31-08-117_com aniplex fategrandorder](https://user-images.githubusercontent.com/42795054/217398251-3f301b54-88d2-4ae2-ac33-1bc3ed1630b4.jpg)

vybze:
Why didn't you post images for all three skills?

imniwa:
> Why didn't you post images for all three skills?

all three skills have the same option. later I will post with videos

imniwa:
https://drive.google.com/file/d/1E5GhWyp3125wlayzsryryn53ykfdowwc/preview

chillooo:
This skill confirmation mode could be a new trending, especially those servants have quick NP.

PR #1542, opened 2023-02-23
Added skill options for Kukulkan

squaresmile:
- Added buttons to select option 1 or option 2 for Kukulkan
- For Kuku's 2nd skill, select a target after selecting option
  - Added new skill target syntax `(<target1><target2>...)`. For example Kuku 2nd skill, consume 10 stars and select the first servant would be `b(U1)`

Fixed https://github.com/Fate-Grand-Automata/FGA/issues/1532


app/src/main/res/values-b+zh+CN/localized.xml
```diff
@@ -65,6 +65,9 @@
     <string name="skill_maker_quick">迅击</string>
     <string name="skill_maker_arts">技击</string>
     <string name="skill_maker_buster">力击</string>
+    <string name="skill_maker_kukulcan">Kukulcan</string>
```
reconman, line 68: Don't use English values in other languages, ignore the warnings about missing translations instead.

The app will display the English values if the translated ones are missing.

app/src/main/java/com/mathewsachin/fategrandautomata/ui/skill_maker/SkillMakerViewModel.kt
```diff
@@ -160,7 +160,7 @@ class SkillMakerViewModel @Inject constructor(
         add(
             SkillMakerEntry.Action(
                 when (skill) {
-                    is Skill.Servant -> AutoSkillAction.ServantSkill(skill, target)
+                    is Skill.Servant -> AutoSkillAction.ServantSkill(skill, if (target == null) null else listOf(target))
```
reconman, line 163: Kotlin offers `listOfNotNull(nullable)`, which either creates an empty list on `null` or creates a read-only list containing the object.

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/models/AutoSkillCommand.kt
```diff
@@ -29,14 +29,38 @@ class AutoSkillCommand private constructor(
             return target
         }
 
+        private fun getTargets(queue: Queue<Char>): List<ServantTarget>? {
+            val targets = ArrayList<ServantTarget>()
```
reconman, line 33: In Kotlin you should use `mutableListOf<ServantTarget>()`

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/models/AutoSkillCommand.kt
```diff
@@ -29,14 +29,38 @@ class AutoSkillCommand private constructor(
...
+                }
+            } else {
+                val target = ServantTarget.list.firstOrNull { it.autoSkillCode == nextChar }
+                if (target != null) {
+                    queue.remove()
+                    targets.add(target)
+                }
+            }
+
+            if (targets.isEmpty()) return null
```
reconman, line 52: It's better to use empty lists than nullable lists.

app/src/main/java/com/mathewsachin/fategrandautomata/ui/skill_maker/SkillMakerKukulcan.kt
```diff
@@ -0,0 +1,65 @@
...
+    goToTarget: Boolean,
+    onTarget: (firstTarget: ServantTarget) -> Unit
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxHeight()
+            .padding(16.dp)
+    ) {
+        Text(
+            stringResource(R.string.skill_maker_kukulcan),
```
reconman, line 29: For some reason, this text is black in dark mode, so you can't read it.

Space Ishtar and Emiya don't have this problem but I can't find out why.

I manually specified `color = onSurface` in `FGATitle` and replaced the `Text()` calls with `FGATitle()`.

app/src/main/java/com/mathewsachin/fategrandautomata/ui/skill_maker/SkillMakerKukulcanTarget.kt
```diff
@@ -0,0 +1,64 @@
...
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.mathewsachin.fategrandautomata.R
+import com.mathewsachin.fategrandautomata.scripts.models.ServantTarget
+
+@Composable
+fun SkillMakerKukulcanTarget(
```
reconman, line 17: Since they are closely related, I moved this function to the same file as SkillMakerKukulcan.

reconman:
Thanks for your contribution. I fixed the issues pointed out above myself, but I can't tell if my commit broke anything.

Can you please test my changes?

scripts/src/main/java/com/mathewsachin/fategrandautomata/scripts/models/AutoSkillCommand.kt
```diff
@@ -29,14 +29,38 @@ class AutoSkillCommand private constructor(
             return target
         }
 
+        private fun getTargets(queue: Queue<Char>): List<ServantTarget>? {
+            val targets = ArrayList<ServantTarget>()
+            val nextChar = queue.peek()
+            if (nextChar == '(') {
+                while (true) {
```
reconman: This is dangerous because if the user manually edits the command (which can be enabled in the options), then they may forget to add a `)` and the queue will run out of elements, leading to a `NoSuchElementException`.

I replaced it with an Exception with a proper error text, but it doesn't have translations since the error shouldn't happen to most users.

squaresmile:
Thank you for the suggestions. I just tested the new apk and it works fine.

PR #1582, opened 2023-05-12
Update zh_CN translations

MetLee:


the3dsandwich:
The translation looks good, mostly were changes from half-width characters to full-width characters.

PR #1622, opened 2023-07-07
Improvements for common user errors

reconman:
- Onboarding screens for battery optimization, support images, skill confirmation, youtube
- Autoremove l command when using order swap afterwards
- Use return button instead of click in top right for Servant info windows

The return button can't be used if there are optional windows since it will ask the user if they want to withdraw from the quest.


reconman:
@item4 @ScathachSkadi @the3dsandwich @MetLee Please translate the changed texts in localized.xml.

@the3dsandwich TW is especially outdated. You'll have to look through the history of the English localized.xml because I also changed some existing English texts like accessibility_disabled_message.

MetLee:
sure

the3dsandwich:
@reconman I'm on TW translation, but I'm not sure what **Thresholded stage counter detection (p_thresholded_stage_counter) **does, would you mind letting me know where I can look for what it does? CN translation mentions using binary pictures to detect stage number, but I'm still kind of confused

reconman:
@the3dsandwich Check here for an explanation: https://github.com/Fate-Grand-Automata/FGA/pull/1243

reconman:
Merging now, will add translations later through separate PRs.

PR #1626, opened 2023-07-10
CN translations

MetLee:


reconman:
@MetLee The preferred way of updating translatings is now via this website: https://poeditor.com/join/project/67PXOyBGI0

There are a few strings that are still untranslated and I'll notify you about any further translation needs via the website.

Issue #1665, opened 2023-08-19
Materials Drop Summary


Spiralem:
### Preparation

- [X] No one has asked for this feature before in other issues (even the closed ones)

### Describe the feature

I am wondering if it is possible to have the Materials drop summary at the end of the script to list Vertically instead of Horizontally as some of the materials usually have long names. (e.g. similar to image attached)
Or if it is to tailor to some devices display?
![Untitled](https://github.com/Fate-Grand-Automata/FGA/assets/26656490/89614348-91b3-4620-8b5b-002619d305bc)

PR #1678, opened 2023-09-09
Change LazyRow to FlowRow for Material Drops

ArthurKun21:
#1665

I think FlowRow would be great for this

[FlowRow](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#FlowRow(androidx.compose.ui.Modifier,androidx.compose.foundation.layout.Arrangement.Horizontal,androidx.compose.foundation.layout.Arrangement.Vertical,kotlin.Int,kotlin.Function1))

Although it is an ExperimentalAPI

https://github.com/Fate-Grand-Automata/FGA/assets/16458204/66ffefef-5655-48e3-af8c-d0b73e2b1407

### DIFF
--- a/app/src/main/java/io/github/fate_grand_automata/ui/exit/BattleExit.kt
+++ b/app/src/main/java/io/github/fate_grand_automata/ui/exit/BattleExit.kt
@@ -226,6 +226,9 @@ private fun MaterialSummary(
         modifier = Modifier.padding(16.dp)
     ) {
         for ((mat, count) in materials.toList()){
+            if (count == 0){
+                continue
+            }
             Card(
                 shape = CircleShape,
                 modifier = Modifier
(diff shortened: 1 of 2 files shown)


reconman:
I'm thinking of only showing any materials which dropped so you can just enable all materials for one generic battle config.

reconman:
I've refactored your change a bit so the MaterialTracker doesn't return empty materials in the getter. That way, we don't need to check each entry in the for-loop.

reconman:
Oh also I would recommend using branches and not working on master for PRs since now you have to force-reset master:

```
git fetch https://github.com/Fate-Grand-Automata/FGA.git
git stash
git reset --hard FETCH_HEAD
git push -f origin master
```

Afterwards, you can apply your stashed changes from the 2nd command again with `git stash pop` if you had any uncommitted changes to files.

PR #1680, opened 2023-09-11
Migrate groovy to kts

ArthurKun21:
Haven't tested the released and CI, but it is still working on my device and haven't found any bugs yet

PR #1683, opened 2023-09-17
Remember No. of apples/runs/mats/CEs/battle config per server, updated Card Priority UI, Ordeal Call out of pods fix, Auto Servant Level, Auto Skill Level

ArthurKun21:
also included the #1680 

- Remembers the selected battle config per server
- Remembers the number of apples inputted(Still need to select `Ok` to it will remember
- Fix the bug with the pager when all items in the page was deleted and you are scrolled to that specific page, it will caused an error as the page rendering is no longer available(now it will scroll to the initial page)


app/src/main/java/io/github/fate_grand_automata/ui/launcher/BattleLauncher.kt
```diff
@@ -59,16 +65,79 @@ fun battleLauncher(
...
-    var limitRuns by remember { mutableStateOf(prefs.refill.limitRuns) }
-    var shouldLimitMats by remember { mutableStateOf(prefs.refill.shouldLimitMats) }
-    var limitMats by remember { mutableStateOf(prefs.refill.limitMats) }
-    var shouldLimitCEs by remember { mutableStateOf(prefs.refill.shouldLimitCEs) }
-    var limitCEs by remember { mutableStateOf(prefs.refill.limitCEs) }
+
+    val perServerConfigPref by remember {
+        mutableStateOf(
+            prefs.perServerConfigPrefList.single { selected ->
+                prefs.gameServer == selected.server
```
reconman, line 72: A NoSuchElementException is thrown here if you use game servers for BetterFGO.

```suggestion
                prefs.gameServer.simple == selected.server.simple
```

reconman (changes requested):
This is a preliminary review, I haven't tested your new apple settings yet.

app/src/main/java/io/github/fate_grand_automata/ui/card_priority/CardPriorityDragSort.kt
```diff
@@ -1,50 +1,182 @@
...
+    onSubmit: (MutableList<CardScore>) -> Unit
 ) {
-    val context = LocalContext.current
+    var cloneScores by remember { mutableStateOf(scores.toList()) }
 
     DragSort(
-        items = scores,
-        viewConfigGrabber = {
-            DragSortAdapter.ItemViewConfig(
-                foregroundColor = Color.WHITE,
```
reconman, line 23: The font is not white anymore so it's hard to read.

app/src/main/java/io/github/fate_grand_automata/ui/card_priority/Render.kt
```diff
@@ -81,39 +102,129 @@ fun CardPriorityListItem.Render(
...
+        },
+        onSubmit = {
+            clonePriorities = it
+            onSubmit(clonePriorities.toMutableList())
         }
     )
+}
+
+@Composable
+private fun PrioritiesCardItem(
```
reconman, line 192: The text is not white anymore for the HIGH -> LOW priority and SERVANT PRIORITY items.

Also, in both lists the items got wider, so on my phone the RQ entry is not on the screen anymore. If I drag the items to the left so I can see RQ, I also automatically scroll to wave 2 priority.

reconman (changes requested):

app/src/main/java/io/github/fate_grand_automata/ui/main/MainScreenViewModel.kt
```diff
@@ -22,6 +26,21 @@ class MainScreenViewModel @Inject constructor(
     val prefs: IPreferences,
     val storageProvider: StorageProvider
 ) : ViewModel() {
+
+    init {
+        // init the perServerConfigs
+        CoroutineScope(Dispatchers.IO).launch {
```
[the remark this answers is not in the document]
reconman: Not 100 % sure it's needed here but I'll leave it in.

app/src/main/java/io/github/fate_grand_automata/util/StorageProvider.kt
```diff
@@ -45,8 +45,12 @@ class StorageProvider @Inject constructor(
     val rootDirName
         get() = dirRoot?.name
 
-    private val recordingFile
-        get() = dirRoot.getOrCreateFile("record.mp4")
+    private val recordingFile: DocumentFile
+        get(){
+            val sdf = SimpleDateFormat("dd-M-yyyy-hh-mm-ss", Locale.US)
```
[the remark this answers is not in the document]
reconman: Then that one should be changed too.

app/src/main/java/io/github/fate_grand_automata/ui/card_priority/CardPriorityDragSort.kt
```diff
@@ -1,50 +1,186 @@
...
 import io.github.fate_grand_automata.R
 import io.github.fate_grand_automata.scripts.enums.CardAffinityEnum
 import io.github.fate_grand_automata.scripts.enums.CardTypeEnum
 import io.github.fate_grand_automata.scripts.models.CardScore
+import io.github.fate_grand_automata.ui.customFGAColors
 import io.github.fate_grand_automata.ui.drag_sort.DragSort
-import io.github.fate_grand_automata.ui.drag_sort.DragSortAdapter
 
 @Composable
 fun CardPriorityDragSort(
```
reconman, line 47: I just tested drag and drop in landscape mode and it's pretty much unusable because you only ever see 2 rows.

I think we should remove it again unless you can come up with some way to make it work in landscape mode.

reconman (changes requested):

Issue #1702, opened 2023-11-09
Blue apple option missing from FGA for TW server


ALIAZ333:
### Preparation

- [X] I tested the [latest release](https://github.com/Fate-Grand-Automata/FGA/releases/tag/canary)
- [X] I looked at other issues (even the [closed ones](https://github.com/Fate-Grand-Automata/FGA/issues?q=is%3Aissue+is%3Aclosed))
- [X] I read the [Troubleshooting Guide](https://github.com/Fate-Grand-Automata/FGA/wiki/Troubleshooting)

### FGO server

TW

### FGA build number

2242

### Describe the bug

Blue apple was recently implemented to TW server. Could you kindly add the option to select blue apples? Thanks a lot!!

### Video

![Screenshot_20231110-032024](https://github.com/Fate-Grand-Automata/FGA/assets/133506428/39ae0b47-7f98-4c5b-b3e3-aecc73a526b3)

### Device model

Galaxy Note20 Ultra 5G

### Android version

11

the3dsandwich:
@ALIAZ333 tagged you in #1703, would you mind seeing if this works for you?
use the apk from https://github.com/Fate-Grand-Automata/FGA/actions/runs/6840149604

ALIAZ333:
Just tested, the apk works just fine! Thanks @the3dsandwich!

PR #1703, opened 2023-11-12
Enable bronze apple in TW

the3dsandwich:
Addresses #1702

Don't know if it's blue of bronze but makes the option appear when selecting battle config


the3dsandwich:
@ALIAZ333 would you mind testing?

ALIAZ333:
Absolutely! Will report back once tested!

PR #1726, opened 2023-12-30
Added Auto screenshot of bond CE and bond level up! And reduce the Play button's alpha to make it more translucent

ArthurKun21:
- added auto screenshot of bond level up(regardless of level)
- added auto hide of the play button before screenshot updated play button's transparency when running
- updated play button's transparency when running

### DIFF
--- a/prefs/src/main/java/io/github/fate_grand_automata/prefs/Preferences.kt
+++ b/prefs/src/main/java/io/github/fate_grand_automata/prefs/Preferences.kt
@@ -82,6 +82,10 @@ class PreferencesImpl @Inject constructor(
 
     override val screenshotDropsUnmodified by prefs.screenshotDropsUnmodified
 
+    override val screenshotBond by prefs.screenshotBond
+
+    override var hidePlayButtonForScreenshot by prefs.hidePlayButtonForScreenshot
+
     override val stageCounterSimilarity by prefs.stageCounterSimilarity.map { it / 100.0 }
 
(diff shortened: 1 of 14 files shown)


prefs/src/main/java/io/github/fate_grand_automata/prefs/Preferences.kt
```diff
@@ -82,6 +82,10 @@ class PreferencesImpl @Inject constructor(
 
     override val screenshotDropsUnmodified by prefs.screenshotDropsUnmodified
 
+    override val screenshotBond by prefs.screenshotBond
+
+    override var hidePlayButtonForScreenshot by prefs.hidePlayButtonForScreenshot
```
reconman, line 87: Preference can't be changed by users right now

reconman (changes requested):

prefs/src/main/java/io/github/fate_grand_automata/prefs/Preferences.kt
reconman: Sorry for the late reply, only read your comments now. I was talking about how `hidePlayButtonForScreenshot` can't be changed.

reconman: I would prefer if the variable had `@ScriptScope` or `@ServiceScoped`, but I think so far preferences are the only thing shared by both which can affect the UI.

PR #1740, opened 2024-01-02
Lottery-Gift Box Loop

ArthurKun21:
- added lottery-gift box loop until full of embers and present box
- While the NA config was added, it can't be 100% tested.
  - TW was added and tested


ArthurKun21:
it has been 2 years, I don't think this is going to be possibly be added

PR #1743, opened 2024-01-04
added Daily Free FP summon to Auto Friend Gacha

ArthurKun21:
added the initial fp summon location


reconman:
I'll test it tomorrow when I get my free 10-pull again, just to make sure I know how it works.

Issue #665, opened 2021-01-10
CE bomb making script


Whatwaq:
Before opening an issue:
- [yes ] No one has asked for this feature before in other issues (even the closed ones)
This script would, after the CE spirit orgin list is full, use 1/2 and optionally 3 star ces to enhance a single ces

Inside the enhance-craft essence menu, FGA would scan all the craft essences for a match of the one to be enhanced, so using the support image maker to capture a image of the ce would be needed. 

> If you open unnecessary/dumb issues, you'll be rewarded with appropriate tags.

MathewSachin:
Duplicate of #550.
I personally would never make this, so closing.

Issue #550, opened 2020-11-06
CE Bomb script


sadmafioso1:
Thanks for making this app free

PR #1749, opened 2024-01-12
Auto CE Level/CE bomb

### LINKED ISSUE #998: Adjust CE Bomb for wide screen, add documentation

ArthurKun21:
- fixes #998
- #733
- #665 
- #550

---

- Supports EN, TW and JP only.
  - PRs for other servers are welcome.
- Added multiple options to do CE Bomb
- Added FP Gacha redirect to ce bomb
  - Will only happen when the CE Storage is full as we got guaranteed it would directly go to ce menu.
- Supports Long press and Drag for faster creation of CE bomb
  - For Android 8 and above only. For Android 7.1 and below, it will just be multiple clicks as long press is not supported in API
- Auto locking of the current target CE
  - As long as you have not disabled it with the option.
  - For Android 8 and above only.
- It should be auto skip locked CEs.
- Please use the smallest display for CEs, this was optimized for that.
- Supports:
  - Target: 1* and 2* only
  - Fodder 1*, 2*, and 3* only.
    - If you want to use 4* CEs You have to disabled the auto filtering and manually click the 4* CEs. The automation for it was deliberate as to not accidentally feed other non EXP 4* CEs.
- You can select your own Target CE, and make the script just select fodders for you, skipping the selection of target CE.
  - [x] For Fix, stop once the own target CE is max level

Auto CE Level
- when user have preselect the target CE. The script would stop either once the target CE is at max level or ran out of fodder CEs.

Auto CE Bomb
- when user don't have preselect target CE. The script would stop once it ran out of fodder CEs.

---


potchy:
Why did you close it? :(

KaitoKid:
@ArthurKun21  can you regenerate another artifact for this build? The previous is expired

ArthurKun21:
it has been 2 years, I don't think this is going to be possibly be added. Especially with the project seemingly in maintenance mode.

PR #1751, opened 2024-01-13
Auto Append

ArthurKun21:
- EN and JP only
  - TW, untested. Also No Insufficient QP template
  - PR for other servers are welcome.
- Unable to read the number behind the text. So users need to manually input the number of upgrades.
- Has the ability to unlock appends.


ArthurKun21:
it has been 2 years, I don't think this is going to be possibly be added. Especially with the project seemingly in maintenance mode.

PR #1775, opened 2024-03-01
Command Code Reward

ArthurKun21:


app/src/main/assets/En/command_code_reward.png
reconman, line 1: I'm a bit worried that the image is too small and there may be false positive matches because of it.

PR #1782, opened 2024-03-26
Early Support Screen Refresh

### LINKED ISSUE #1130: Detect bottom of support list
### Preparation

### Describe the feature

Is it possible to add a "scroll to bottom" feature in selecting support list? I believe this could be done just by checking the bottom pixels of the scroll bar (i.e. if it's white or not). This way, no extra time would be spent after scrolling to the end of the list.

image

ArthurKun21:
Fixes #1130

- Hit Refresh if it is available instead of waiting(usually happens on initial support check)
- When the scrollbar hit the bottom, perform refresh instead of doing succeeding swipes
- If there is no scrollbar, after checking the servants and if no matched then perform refresh


reconman:
Thanks, wanted to implement something for a long time but never did.

PR #1792, opened 2024-04-04
Lotto out of currency

ArthurKun21:
- Moved the lotto click which causes the failure to detect the out of currency
- Removed the legacy detection of out of currency, added a new redundant checker to ensure it is really out of currency
- Increase the delay between lottery and Gift Box, due to detecting the gift box while the gift box is still not on screen


scripts/src/main/java/io/github/fate_grand_automata/scripts/entrypoints/AutoLottery.kt
```diff
@@ -39,33 +39,13 @@ class AutoLottery @Inject constructor(
...
-        spin()
-        if (isNewLineup()) {
-            confirmNewLineup()
-        } else if (isOutOfCurrency()) {
-            throw ExitException(ExitReason.RanOutOfCurrency)
-        }
-    }
-
-    private fun isNewLineup() =
-        images[Images.LotteryLineupUpdated] in locations.lottery.lineupUpdatedRegion
```
reconman, line 63: Please remove the LotteryLineupUpdated image if it's no longer needed.

reconman (changes requested):

Issue #1100, opened 2022-02-21
Getting stuck on party selection


Kotsin:
### Preparation

- [X] I tested the [latest release](https://github.com/Fate-Grand-Automata/FGA/releases/tag/canary)
- [X] I looked at other issues (even the [closed ones](https://github.com/Fate-Grand-Automata/FGA/issues?q=is%3Aissue+is%3Aclosed))
- [X] I read the [Troubleshooting Guide](https://github.com/MathewSachin/Fate-Grand-Automata/wiki/Troubleshooting)

### FGO server

NA

### FGA build number

1611

### Describe the bug

When the app is trying to start the battle, it's getting stuck on the party selection screen. I know that it's an old bug in FGO and the app is supposed to handle it, but it doesn't. 

### Video

https://drive.google.com/file/d/1zBR6Ub0J1BJoMH-F2mrLch6SWoiSNhZA/view?usp=sharing

### Device model

Poco X3 Pro

### Android version

11

### Screen size

2400x1080

### RAM

8

Puh00:
I also faced this issue when I used the emulators Nox and LDPlayer

NoOneofConsequencePrime:
I found a temporary fix to this by clearing "party selection" to default, without the click after selecting support the battle seems to start properly (just remember to start with the correct party).

BrutuZ:
Suggestion: FGA could work around this by going back and forth 1 party.

- Tap chosen party # 
- Tap right corner to go to next party
- Tap left corner to go to back to chosen party

reconman:
@BrutuZ we already used that logic once but changed it to an image detection one later. Right now the bug only happens on a few devices where I don't know why.

BrutuZ:
It's happening on BlueStacks 5 for me. Now I'm curious if it'd happen on my stand-alone device.

The way I see it, it's as OP pointed out, a FGO bug, not FGA's fault, it can be reproduced even if FGA isn't running, just tap any party dot and the start mission button becomes unresponsive. FGA could still workaround it the way we usually do when using the _skip to party_ dots manually.

PR #1817, opened 2024-05-20
Fix JP Template image

ArthurKun21:
fix by just using the EN template


reconman:
Actually, I think we need to cut down the image to just the yellow pixels to fix #1100 

The image changes from black to white background after clearing Solomon.

reconman:
Replaced with [25af12b](https://github.com/Fate-Grand-Automata/FGA/commit/25af12b16b16a430f0218e86a6f6ec29d69d2c46)

PR #1832, opened 2024-06-01
Select Servant with Append in Support screen 

ArthurKun21:
- Added additional checking for appends, entirely optional and will be skip if the user didn't select it
  image

- Upon checking of this feature turns out that there are 3 display states for skill display checking
  1. Skill
  2. Append
  3. Switching Skill & Append every few seconds
 - In order to lessen the time it takes for checking of this state:
   - Assume the initial state was `Skill display`, perform no checking of the skill state
   - Upon checking the append, just make sure it was not on the `skill display` state and if it was, then press the skill display button to go into the `append display`
   - After the checking of the Appends, just click the Skill display twice to return to the `Skill display` state
 - Perform no checking of append if the follow was on the right side, this is due to the fact that story/event servants don't have append
   story-no-append
- Don't have images for other servers, so just going to assume and perform no checking, YOLO!


Cheesecake-D:
@ArthurKun21 I’m new to this app and srry if this sound stupid but can I ask where can I find the also check append skill?

PR #1846, opened 2024-06-09
Improved Battle Config List Selection UI

ArthurKun21:
- Removed Check mark for selection replaced with changing background
- Added double click to start selection mode


reconman:
I don't like the UI change to the selection mode and I don't think double tapping is intuitive, closing this.

ArthurKun21:
it was common ux pattern to distinguished selected items, and the current check mark selected item UI lags when there are too many items as it will redraw all of the items in the list when it enters selected mode

PR #1851, opened 2024-06-25
No media

ArthurKun21:
- added nomedia on the individual support servant folder
- added nomedia on the bond screenshot folder
- added nomedia on the dump folder


reconman:
.nomedia is only needed in the root directory of FGA, not in every subdirectory.

I haven't looked at the current logic yet. I tested it at some point with a new directory and it worked back then.

reconman:
I can see that the bond directory images show up, but the support images don't.

When I extract the default support images, the .nomedia file gets created in the support folder and all support images are hidden in my gallery.

PR #1875, opened 2024-08-12
App Permission

ArthurKun21:
Fix for google play by removing the location in the foreground as we don't need it.


reconman:
Has nothing to do with foreground service location.

PR #1926, opened 2024-11-27
Switch from tesseract to google text-recognition

reconman:


ArthurKun21:
need to push this or update tesseract in a while

https://github.com/Fate-Grand-Automata/FGA/actions/runs/14648963323/job/41109958586

reconman:
The APK didn't get any smaller so I didn't continue with this approach.

reconman:
Tested Google ML text recognition again. It doesn't work with the current code for numbers like x6 where it returns an empty string instead.

PR #2054, opened 2025-07-30
Fix auto-leveling for JP

### LINKED ISSUE #2053: Auto leveling is broken after 10th anniversary update in JP
### Preparation

- [x] I tested the [latest release](https://github.com/Fate-Grand-Automata/FGA/releases/tag/canary)
- [x] I looked at other issues (even the [closed ones](https://github.com/Fate-Grand-Automata/FGA/issues?q=is%3Aissue+is%3Aclosed))
- [x] I read the [Troubleshooting Guide](https://github.com/Fate-Grand-Automata/FGA/wiki/Troubleshooting)

### FGO server

JP

### FGA build number

2982

### Describe the bug

JP introduced a few new UI elements in the Servant Enhancement screen.

New "Clear" button in the top right and the Auto Select button was moved to the left.
Image

New Automatic Set toggle under the rarity selection, which automatically fills the 20 slots with Embers after the enhancement was done.
Image

They also added automatic CE leveling to this update.

Image
Image

### Video

https://github.com/user-attachments/assets/272d2f5f-6a11-4812-b8fa-75dadbe13f35

### Device model

Not specified

### Android version

15

### Screen size

_No response_

### RAM

_No response_

reconman:
Fixes #2053

For now, it only moves the region of the Auto Select button to the left for JP and the images were cropped to ignore the 自動セット text.

That way, it should probably be able to handle both variants, ON and OFF. For the ON situation, it clicks on Auto Select and OK even though Embers are already selected, but I didn't want to maintain 4 images per language instead of 2 and it only wastes a few seconds per loop.


reconman:
Open issue: the dialog for the ember selection filters said 決定 in the past, now it's 実行

PR #2061, opened 2025-08-09
fix consumption of find all images twice

ArthurKun21:
This pull request updates the `findAll` method in the `ImageMatcher` interface and its implementation to improve reliability and consistency by changing the return type from `Sequence<Match>` to `List<Match>`. This change ensures that match results are fully realized and avoids potential issues with sequence consumption and side effects.

**Interface and implementation changes:**

* Changed the return type of `findAll` in the `ImageMatcher` interface from `Sequence<Match>` to `List<Match>` for more predictable and safer usage.
* Updated the `findAll` implementation in `RealImageMatcher` to collect matches into a list before returning, preventing issues related to consuming sequences multiple times. [[1]](diffhunk://#diff-bcacdb6c8bf30ae17217371cb51709a41e5075bdbc03e2b2c5b26f370ee42507L147-R148) [[2]](diffhunk://#diff-bcacdb6c8bf30ae17217371cb51709a41e5075bdbc03e2b2c5b26f370ee42507L162-R169)

### DIFF
--- a/libautomata/src/main/java/io/github/lib_automata/ImageMatcher.kt
+++ b/libautomata/src/main/java/io/github/lib_automata/ImageMatcher.kt
@@ -46,7 +46,7 @@ interface ImageMatcher {
         region: Region,
         pattern: Pattern,
         similarity: Double? = null
-    ): Sequence<Match>
+    ): List<Match>
 
     fun isWhite(region: Region): Boolean
     fun isBlack(region: Region): Boolean
@@ -144,8 +144,8 @@ class RealImageMatcher @Inject constructor(
         )
     }
 
-    override fun findAll(region: Region, pattern: Pattern, similarity: Double?) =
-        screenshotManager.getScreenshot()
+    override fun findAll(region: Region, pattern: Pattern, similarity: Double?): List<Match> {
+        val matches = screenshotManager.getScreenshot()
             .crop(transform.toImage(region))
             .findMatches(
                 pattern,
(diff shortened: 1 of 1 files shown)


ArthurKun21:
This should halved our image matching processing calls.

The issue was `highlight` calls the sequence all over again even though the operation was already completed once.

reconman:
What exactly is the error you've noticed?

From what I can see, `SequenceBuilderIterator`, which is created in `findAll()` only checks if there is a result in `.hasNext()`, but does not consume the `Match` object.

ArthurKun21:
if you check the logs you can actually see that it is consuming twice but the 2nd one wasn't supposed to be called. Turns out the `also` + `sequence` makes it called it twice so I split the two of them

```
DroidCvPattern$findMatches: Stopped matching En/GoldXP.png at score (0.38154539465904236) < similarity (0.8)
DroidCvPattern$findMatches: Stopped matching En/GoldXP.png at score (0.38154539465904236) < similarity (0.8)
DroidCvPattern$findMatches: Stopped matching En/SilverXP.png at score (0.3556382656097412) < similarity (0.8)
DroidCvPattern$findMatches: Stopped matching En/SilverXP.png at score (0.3556382656097412) < similarity (0.8)
DroidCvPattern$findMatches: Stopped matching En/Gold5StarXP.png at score (0.4151449501514435) < similarity (0.8)
DroidCvPattern$findMatches: Stopped matching En/Gold5StarXP.png at score (0.4151449501514435) < similarity (0.8)
```

https://github.com/Fate-Grand-Automata/FGA/blob/fa6732bdb7e9b4b96d3ead8ec287c82c5bcdb447/scripts/src/main/java/io/github/fate_grand_automata/scripts/entrypoints/AutoDetect.kt#L27-L28

PR #2065, opened 2025-08-10
feat: add support for Grand duels support selection

erectorOps:
This pull request adds the following features to the support selection screen for Grand Duel (Grand Coronation Battle) on the JP server.
・Added an option to match only Grand Class in Servant image matching.
・Added an option to match based on how many registered images match in Craft Essence image matching.
・Made Support Image Maker compatible with Grand Servant layouts.
This text was created using DeepL's Japanese-English translation service.


erectorOps:
-  This is an example of matching only with Altera of Grand Saber.
guide1
<br><br>

- This is an example of matching only with support equipped with both “Chaldea Lunchtime” and “Chaldea Teatime.”

guide2

ArthurKun21:
grand_servant_ce_icon

just use the simple icon to become server agnostic

Hayantes:
works fine for me, huge thanks for work !

1egenDary:
Finally! Thanks a lot

vybze:
Screenshot_20250523-152156
Screenshot_20250523-152940
Match icons for bond ce effects instead of having to create a third ce for matching

reconman:
> Screenshot_20250523-152156 Screenshot_20250523-152940 Match icons for bond ce effects instead of having to create a third ce for matching

So in this case, only the top and bottom CE should be checked against the CEs recorded via Support Image Maker. The center one should be checked if the Bond CE effect or the NP gain effect was chosen.

Since the NP gain icon is pretty small, I think it's enough to check for this image:
Bond CE Effect

If this icon cannot be found, the friend has configured the bond CE to give NP charge.

erectorOps:
> Since the NP gain icon is pretty small, I think it's enough to check for this image:
> If this icon cannot be found, the friend has configured the bond CE to give NP charge.

Since there are cases where the 'Bond CE' is not equipped, it needs to be checked twice.

erectorOps:
Added items for Bond CE effect icons to the Preferred CEs list.
These items will always exist regardless of the Support Image folder.
Screenshot_20250813-130343

vybze:
> Added items for Bond CE effect icons to the Preferred CEs list. These items will always exist regardless of the Support Image folder. Screenshot_20250813-130343

Probably gonna need to add a notice for when someone uses the match 3 option that they need to select a bond ce effect or else they'll lock fga out of selecting a support

erectorOps:
Removed extra options from the Preferred CEs and reorganized related settings for clarity.
(Since I’m not a native English speaker, I’m not sure if the result is truly clear — feel free to suggest improvements.)
Screenshot_20250814-022455

vybze:
> Removed extra options from the Preferred CEs and reorganized related settings for clarity. (Since I’m not a native English speaker, I’m not sure if the result is truly clear — feel free to suggest improvements.) Screenshot_20250814-022455

The bond ce preference is fine but the normal and rewards slot will have to change. I'll look up the proper english names later when I have some time

reconman:
I think "Bonus Reward (Bond/FP/QP/EXP)" would be a better term instead of just "Reward".

In Japanese it's called "報酬アップ" (Reward Up), so the 2nd part was missing in the English translation.

I would add the "(Bond/FP/QP/EXP)" part to all languages so it's easier to understand what "Reward" means.

erectorOps:
I updated the wording based on the advice received.
Please see the attached screenshot for the updated screen.
Screenshot_20250814-053354

vybze:
I would go with free slot or any ce slot and bonus up (bond/fp/qp/exp) slot. Sounds better in my opinion

erectorOps:
I sincerely appreciate everyone who took the time to review this.
Your thoughtful feedback was invaluable in refining these changes.
I believe the PR is now well-polished and ready for final approval.
Please feel free to let me know if there’s anything else.

PR #2067, opened 2025-08-13
chore: move the file paths to Images for easier management

ArthurKun21:
I move the path to Images for easier management.

Need some verification before merge


ArthurKun21:
This pull request refactors the way image file paths are associated with the `Images` enum, making the code more maintainable and easier to extend. The key change is to move the mapping of enum values to file names directly into the enum as a property, eliminating the need for a separate mapping function.

**Refactoring and Simplification:**

* The `Images` enum in `Images.kt` now includes a `path` property, with each enum value directly associated with its corresponding file name. This replaces the previous approach where a separate `fileName()` function handled this mapping.
* The `ImageLoader` class in `ImageLoader.kt` is updated to use the new `path` property from the `Images` enum instead of calling the removed `fileName()` function, simplifying image path access throughout the codebase.

Issue #2073, opened 2025-08-17
求助：在能放就放模式里，FGA会点击尚在冷却中的技能，并导致卡住


xuejianxianzun:
### Preparation

- [x] I read the [Troubleshooting Guide](https://github.com/Fate-Grand-Automata/FGA/wiki/Troubleshooting)
- [x] My question isn't answered in any other issues (even the closed ones)

### Your question

我尝试过手动设置技能释放目标，也尝试过 SPAM ALL（没有进行其他设置），但都出现了相同的问题：
在我方的第三轮或更多行动里，FGA 会点击尚在冷却的技能，导致出现技能确认窗口。之后它会点击别的地方让确认窗口消失，但接下来会点击角色头像，并无限卡住。
如果我手动点击空白区域，让界面回复到干净的状态，依然会重复上面的问题（点击技能），然后一直卡住。
使用不同从者都会出现这个情况。当使用的从者不同时，FGA 点击的技能也会变化，比如有时它会点击第一个从者的技能，有时会点击第三个从者的技能，并出现同样的问题。

**Video：**
https://pixeldrain.com/u/qXBno1CM

**Screenshot：**

<img width="1417" height="797" alt="Image" src="https://github.com/user-attachments/assets/5460b8ab-5b78-4226-a389-7bc82dbf5b45" />

<img width="2362" height="1328" alt="Image" src="https://github.com/user-attachments/assets/61a75686-0045-4a8d-bc54-7a010d33e913" />

<img width="2362" height="1328" alt="Image" src="https://github.com/user-attachments/assets/6f3ac807-a755-424f-a739-0ec468a6a5ca" />

<img width="2362" height="1328" alt="Image" src="https://github.com/user-attachments/assets/30414f5a-f4d3-41f5-80e2-404bf9913dd4" />

vybze:
We've gotten a few reports of spam not working has intended, I recommend that you setup a command script instead of using spam

xuejianxianzun:
> We've gotten a few reports of spam not working has intended, I recommend that you setup a command script instead of using spam

Thanks for the reply, that is to say do I need to manually set the timing of the release skill?It's not flexible enough (when the lineup changes), but if there is no other option, it can only be so.

xuejianxianzun:
Thanks for the fix above, it doesn't get stuck now and works fine. 😊

ArthurKun21:
let's wait for the issue to get merge and let it auto close this. So that other people will know about the same issue

[...]

PR #2076, opened 2025-08-19
fix: Skill window not closing during cooldown

erectorOps:
# Fix skill window not closing during cooldown

- [x] Bug Fix  
- [ ] Feature Implementation  
- [ ] Documentation  
- [ ] Other  

## Related Issue  
Partially addresses #2073

## Description  
This PR fixes part of the issue described in #2073 — specifically, a bug where the skill window could not be closed while a skill was on cooldown.  
Other related problems mentioned in the issue may still need to be addressed in future updates.

## Screenshots
screenshot_skill_window_stuck

## Testing  
1. Manually set the target skill into a cooldown state beforehand.  
2. Configure FGA to select and use that skill.  
3. Run the script and observe the behavior during the cooldown.  
4. Confirm that the skill window now closes correctly and no longer remains stuck.


ArthurKun21:
@reconman feels like it would be better to invite him to the repo as Collaborator so he can contribute directly to do the repo. wdyt?

PR #2082, opened 2025-08-20
Downgrade OpenCV to fix crashes

reconman:
User on Discord reported a crash on Bluestacks when matching images and there is also another new crash in the Play Console when converting screenshots to RBG, mostly on Android 15 devices.


ArthurKun21:
weird, I'm on bluestacks and haven't seen a crash with it

android 15 again...

vybze:
Will also add that the user was using bluestack air on Mac

PR #2084, opened 2025-08-21
chore: use kotlin uuid

ArthurKun21:
## Pull Request Type

- [X] Chore
- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Use native kotlin uuid


reconman:
Why do you want FGA to use an experimental kotlin feature that does the same thing as a Java feature that exists for at least 10 years?

ArthurKun21:
To promote the usage of kotlin first

reconman:
From what I can see, both implementations do the same thing (create a byte array and then fill it with values generated by SecureRandom) and switching provides no benefit.

PR #2183, opened 2026-02-17
feat: Split APKs

ArthurKun21:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

- Enable to lower the APK file size of other build variants

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -79,48 +79,45 @@ jobs:
 
           echo "FGA_BUILD_NAME=$build_name" >>${GITHUB_ENV}
 
-          echo "FGA_APK_ARTIFACT_NAME=FGA-apk-PR-$build_name" >>${GITHUB_ENV}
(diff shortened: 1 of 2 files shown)


ArthurKun21:
`pull_request_target`

.github/workflows/pr-build.yml
```diff
@@ -24,9 +24,13 @@ env:
 
 jobs:
   build:
-    runs-on: macos-15
+    runs-on: macos-latest
```
danilaml, line 27: Huh. Why does CI run on macos runners? It doesn't even produce Apple artifacts.

ArthurKun21: faster

danilaml: @ArthurKun21 did something change recently? macos runners were always slower than the linux ones.

ArthurKun21: last time we check, macos was faster for android builds. actions is free anyways

Issue #874, opened 2021-09-11
Adjust Skill regions to fix Skill spam


MathewSachin:
NA skill icons seem to have shifted a bit. Skill spam isn't getting the exact location where skill cooldowns appear, so is not working consistently. Have to update the regions.

**Note:** Skill spam was only ever calibrated for JP and NA. I personally can't do that for other servers.

Issue #540, opened 2020-10-28
Skill Spam Seems to sometimes skip skills.


Rajvir8:
Overall Skill Spam just isn't triggering at times

![Screenshot_20201028-093612](https://user-images.githubusercontent.com/17600697/97443182-19983800-1901-11eb-9087-f3f451a38e06.png)

I was watching and paused it when it saw all my skills and skipped past using them this turn.

The auto I was using to test this out
abcdefgh1i456

Overall it does remember at times, so it's definitely useful on to add to stall team, but it's probably best to still just plan everything out manually as usual and then after it's over it might use some skills on Cooldown.

Overall this remains a great new feature that sadly cannot be relied on as a primary feature just yet.  

Works great as a add-on though.

Also just noticed this new version perfectly avoids stunning cards and I really appreciate that for the record.

MathewSachin:
I had encountered it not using skills before but didn't pay much attention since others said it worked fine and I mostly 3T.
I adjusted the skill check regions and added delays in between in case it was checking for skills before they were visible after the animation.

Try the APK from this page: https://github.com/MathewSachin/Fate-Grand-Automata/actions/runs/333852920

MathewSachin:
Ah, there's one more case to thing to take care of. I'll post a new build after a workaround.

MathewSachin:
New build: https://github.com/MathewSachin/Fate-Grand-Automata/actions/runs/333946957

Rajvir8:
Downloaded the new build, I'll try it out later today when I have so more AP.

Rajvir8:
Actually you know what I'll go do one Salem run for 20AP.  Best place to do it since it's all Casters and Tamamo for my team.

Rajvir8:
It should still be Version 1,087 right?

Nox annoyingly doesn't tell me when it updates it by dragging APK so I always have to wonder if it actually worked.

Rajvir8:
No it did change to 1,093 after I opened it.  Just had a delay and I also manually added it through options as well as just dragging it.

reconman:
1094 also exists, you can see the builds here https://github.com/MathewSachin/Fate-Grand-Automata/actions?query=event%3Apush

Rajvir8:
Yea I'll add that to my notes and swap to that now so I can better test this.

Rajvir8:
[...]

PR #2195, opened 2026-03-16
fix: Add OCR cooldown guard to prevent skill spam false positives

DCPMA:
## Problem

Skill spam occasionally fires skills that are still on cooldown. This happens because the readiness check uses a small 30×30 icon match (`imageRegion`). When a cooldown number is displayed over the skill button, thin glyphs like **1** and **4** don't occlude enough pixels in that sample area, so the icon match still passes the similarity threshold and the skill is cast incorrectly.

This has been reported as a recurring problem (see #2073, #874, #540). Note: #2076 addresses the related "stuck window" symptom (wrong close-click coordinate); this PR addresses the separate root cause of the incorrect cast itself.

## Related issue
- #2073
- #874
- #540

## Related pull request
- #2076

## Fix

Add a secondary OCR guard that reads the cooldown number region of each skill button. If a numeric cooldown is detected, the cast is suppressed regardless of what the icon match says.

**`BattleScreenLocations`** — adds `cooldownTextRegion()` targeting the area where the cooldown number text is rendered. The region is offset from the skill `locate()` point.

**`SkillSpam`** — replaces the direct `in imageRegion` check with `isReadyForSpam()`, which:
1. Takes a snapshot and checks both icon match **and** that no cooldown digit is present (via OCR)
2. If that passes, waits 100 ms and re-checks (stability gate to avoid frame-boundary false positives)
3. Only then allows the cast

The OCR step normalises common Tesseract misreads (`O`→`0`, `I`/`l`→`1`) before extracting the number with a `\d+` regex.

## Scope

- Changes are isolated to `SkillSpam` and `BattleScreenLocations`
- Normal command-script skill casting is unaffected
- No new settings or preferences added

## Testing

Verified on TW server via BlueStacks with a debug build.

### DIFF
--- a/scripts/src/main/java/io/github/fate_grand_automata/scripts/locations/BattleScreenLocations.kt
+++ b/scripts/src/main/java/io/github/fate_grand_automata/scripts/locations/BattleScreenLocations.kt
@@ -117,7 +117,7 @@ class BattleScreenLocations @Inject constructor(
         Region(22, 28, 30, 30) + locate(skill)
 
     fun cooldownTextRegion(skill: Skill.Servant) =
-        Region(8, 108, 96, 48) + locate(skill)
+        Region(-5, 40, 100, 55) + locate(skill)
 
     val servantDetailsInfoClick = Location(-660, 110).xFromCenter()
     


DCPMA:
Closing — the underlying issue (skill spam false positives during cooldown) is still persisting despite this approach. Will revisit with a different strategy.

PR #2196, opened 2026-03-17
feat(battle): custom command card selection

### LINKED ISSUE #2130: Select a specific card sequence in the command line
### Preparation

- [x] No one has asked for this feature before in other issues (even the closed ones)

### Describe the feature

With the advent of servants like Ciel and Olga Marie, strategies that involve forcing all five of a servant's face cards have become more common.

Usually standard card priority works fine, but this can be cumbersome in one wave fights. For instance, on turn one I might want to go NP -> Quick -> Arts, then on turn two go NP -> Buster -> Buster. Having the ability to set different prioirites by turn, or better yet, to simply write what cards to use in the command line, would be a huge help!

Thank you for taking the time to read this.

### LINKED ISSUE #2139: Better Card/Servant Priority
### Preparation

- [x] No one has asked for this feature before in other issues (even the closed ones)

### Describe the feature

**Suggestion:** Add an option for card and servant priority customization as a separate function. Users could choose whether to use the current or the new suggested version. This will allow them to maintain clear speeds and allow for a simpler user experience while allowing for more complex party formation.

**Feature Implementation:** Instead of reading the card type (quick, arts, buster) or the servant (S1, S2, S3); you do both. Then you identify each card as a pair of these two identifiers.

Example: Instead of reading the 5 cards as Q, Q, A, B, Q or S1, S2, S2, S1, S3. You would identify them as Q1, Q2, A2, B1, Q3.

Then you allow the user to take these 9 possible cards and arrange them by priority based on which turn they are on. Give them three turns to choose priorities for.

**UI Implementation:** A simple change from the card priority screen, changes the Q, WQ, RQ, WB... etc to be Q1, Q2, Q3, B1... etc. Then duplicate it twice and label them turns. Add a switch up top to turn on advanced card priority like you have for servant priority. Should allow for minimum work.

**Why This Is Needed:** Some team comps require hitting with a servant's face card to get a little NP gain before the next wave because your refund fell a few percentages short. Or if you wanted to use Calamity Jane's third skill for NP gauge you might need to hit a quick chain to be sure you had the stars come wave 3. But what if you needed both to gen 6% NP gauge and get 50 stars? You can't prioritize card types and a servant. This is for all cases where you need to prioritize different things on different waves or for when you need very specific card combinations and orders to get a little NP. Sure there's probably better options, but not everyone has those options available, or wants to use them. I for one make every effort to use specific servants I need to bond level for coins.

**The Downsides:** This is almost surely going to be slower for the app to run than the current way, instead of checking card type OR servant it has to do both every time the cards change (assuming it already doesn't check both). To prevent this I'm suggesting to add it as a separate option. They hit a switch like for Servant Priority that changes the function called. Maybe a little disclaimer that it's slower if it proves to be in any significance.

sheng-28:
## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related Issue
- Fixed #2139 
- Fixed #2130 

## Description
- Implemented turn-based custom command card selection 
  - Can select up to 3 command cards for each turn
  - Cards defined by CardType and FieldSlot(e.g. B1: Buster of the servant at the front on the field)
  - When all selected cards are available, they would be selected in defined order and card priority-based sort would be skipped
  - Would fall back to card priority sorting when not all selected cards are available
  - When less than 3 cards are selected, more cards would be selected from the rest automatically(e.g. Given user's selection [Q1, A1] and available cards [Q3, B1, A1, B2, Q1], the app would click on [Q1, A1, Q3])
- Added UI on config screen
- Tested on JP server in BlueStacks Air
  - Functions as intended in **Daily Quest** when using _U-Olga's_ or _Ciel's_ shuffle skill at different waves/turns
  - Works fine with _Ishtar_ in **Grand Duel** 
  - Tested with different **Cards Before NP** setup
  - Compatible with **Order Change** (Usually seen in teams that use _Ciel_ + _BB summer_)
- Might need translations

## Screenshots
**Entry on Config Screen with preview**
image
\
**Selection interface designed after card priority interface**
image
\
**Servants ordered as their position on battle field, can select one each time**
image

\
**Up to 3 cards can be selected each turn**
image

### DIFF
--- a/app/src/main/java/io/github/fate_grand_automata/di/script/PreferencesModule.kt
+++ b/app/src/main/java/io/github/fate_grand_automata/di/script/PreferencesModule.kt
@@ -12,6 +12,7 @@ import io.github.fate_grand_automata.scripts.prefs.IPreferences
 import io.github.fate_grand_automata.scripts.prefs.ISupportPreferences
 import io.github.fate_grand_automata.scripts.prefs.ISupportPreferencesCommon
 import io.github.lib_automata.dagger.ScriptScope
+import io.github.fate_grand_automata.scripts.models.CustomCardSelectionPerTurn
 
 @Module
 @InstallIn(ScriptComponent::class)
@@ -50,4 +51,10 @@ class PreferencesModule {
     @Provides
     fun provideServantPriority(battleConfig: IBattleConfig): ServantPriorityPerWave? =
         if (battleConfig.useServantPriority) battleConfig.servantPriority else null
+
+    @ScriptScope
+    @Provides
+    fun provideCustomCardSelection(battleConfig: IBattleConfig): CustomCardSelectionPerTurn =
+        CustomCardSelectionPerTurn.of(battleConfig.customCardSelectionRaw)
+//    todo:replace String config after UI implementation
 }
--- a/app/src/main/java/io/github/fate_grand_automata/di/script/PreferencesModule.kt
+++ b/app/src/main/java/io/github/fate_grand_automata/di/script/PreferencesModule.kt
@@ -55,6 +55,6 @@ class PreferencesModule {
     @ScriptScope
     @Provides
     fun provideCustomCardSelection(battleConfig: IBattleConfig): CustomCardSelectionPerTurn =
-        CustomCardSelectionPerTurn.of(battleConfig.customCardSelectionRaw)
-//    todo:replace String config after UI implementation
+        battleConfig.customCardSelection
+
 }
--- a/app/src/main/java/io/github/fate_grand_automata/ui/battle_config_item/BattleConfigScreen.kt
+++ b/app/src/main/java/io/github/fate_grand_automata/ui/battle_config_item/BattleConfigScreen.kt
@@ -160,6 +160,10 @@ private fun BattleConfigContent(
                             config.notes.EditTextPreference(
                                 title = stringResource(R.string.p_battle_config_notes)
                             )
+                            // todo: remove this after UI implementation
+                            config.customCardSelectionRaw.EditTextPreference(
+                                title = "Custom Card Selection"
+                            )
                         }
                     }
                 }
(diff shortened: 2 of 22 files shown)


ArthurKun21:
Could you also update the wiki with the new entry?

Around here

https://github.com/Fate-Grand-Automata/FGA/blob/master/wiki/scripts/Battle.md#card-priority-optional

sheng-28:
Added an entry in wiki. Not sure if it's clear enough, let me know if I missed anything.

btw, I didn't notice there's a wiki page until today, would've been a great help for me when I started using the app. I wonder why there isn't an in-app button/link to the wiki page?

ArthurKun21:
>  I wonder why there isn't an in-app button/link to the wiki page?

lmao, yea forgot. But personally would love to just have a github pages though

vybze:
How would this work with #2087 ?

ArthurKun21:
> How would this work with #2087 ?

Personally I think, that is out of scope of this PR when the other PR haven't been merged nor reviewed by reconman yet

sheng-28:
> How would this work with #2087 ?


I didn't know abt the other PR so I only had a quick look. Correct me if I'm wrong. 

Based on the implementation of _pickCards()_ in their [Card.kt](https://github.com/Vylantze/FGA/blob/ae70eea515957a9ebbc0a7413770bbb735b7defd/scripts/src/main/java/io/github/fate_grand_automata/scripts/modules/Card.kt) and the one in [Card.kt](https://github.com/sheng-28/FGA/blob/87fef48487dbceb61fdd406b5cfe6d4f6a73ce9a/scripts/src/main/java/io/github/fate_grand_automata/scripts/modules/Card.kt) here, (if both were merged successfully) I think the custom selection feature here would have the highest priority when there is a selection setting and those cards are available, otherwise it would fall back to their implementation.

ArthurKun21:
Let's worry about the other PR once it got reviewed by reconman and merged. The simpler this PR be, the more faster this could get merged.

danilaml:
Seems useful (and I'd imagine it will get even more useful as more servants with card locking mechanics get added to NA). Is it waiting on review from @reconman (like the [other](https://github.com/Fate-Grand-Automata/FGA/pull/2087) tangentially related MR)?

PR #2216, opened 2026-04-28
ci: revert pull_request_target and constraint PR comment to only in-repo

ArthurKun21:
- reverts fc9fd2be0e032a6d678b87172ddd0133ac262865 to only use pull_request and add constraint to only comment when in-repo
- Stacked with #2214

https://www.paloaltonetworks.com/blog/cloud-security/trivy-supply-chain-attack/


reconman:
I've split the workflow into 2 parts per recommendation, PR build and PR comment, the latter with `pull_request_target`.

Took a while to get everything right, but it's working now.

PR #2242, opened 2026-07-12
Fixes bug where there is no valid CE found when there is a valid CE present

ThinkBuildMake:
# Summary
This bug occurs because we don't scroll to the top before trying to find a base CE. Here's what currently happens:

1. Script clicks on "CE to enhance"
2. Since the CEs are sorted by level, level 1 CEs are present at the top
3. We then select the CEs to use for enhancing and enhance the base CE
4. We click CE Enhance again
5. We search for another level 1 base CE WITHOUT SCROLLING TO THE TOP

Here's the problem: If we don't scroll to the top, the UI automatically centers around the original enhanced CE. Since the original enhanced CE is now around level 10 - 15, the UI centers around the bottom. Thus, as CEs build up, there is a chance that no level 1 CEs are found because the UI gets crowded by the already enhanced CEs

For example, in this image, the base CE "Barrier" is enhanced to level 15.
image

As you can see, when it comes time to choose the next base CE, the script wouldn't be able to find it. However, if we were to scroll to the top, we will be able to find some valid CEs to use. 

image

### DIFF
--- a/scripts/src/main/java/io/github/fate_grand_automata/scripts/entrypoints/AutoCEBomb.kt
+++ b/scripts/src/main/java/io/github/fate_grand_automata/scripts/entrypoints/AutoCEBomb.kt
@@ -52,6 +52,10 @@ class AutoCEBomb @Inject constructor(
     class ExitException(val reason: ExitReason) : Exception()
 
     private fun findBaseCE(): Match {
+        // Scroll to top
+        Location(2040, 400).click()
+        2.seconds.wait()
+        
         for (img in imagesForSelectedRarity()) {
             val matches = locations.levelOneCERegion
                 .findAll(images[img])


vybze:
Let me check it out now, you might want to have a look at Arthur's implementation too #1749

ThinkBuildMake:
Thanks for taking a look and for the reference. Let me take a look at it now

ArthurKun21:
> Let me check it out now, you might want to have a look at Arthur's implementation too #1749

It didn't get any review for the past 2 years, I don't think it would be good reference

ThinkBuildMake:
All good. Honestly it's a good reference for me. It definitely does a lot more than I do here. I'm just here to make the script be a bit better so I can be lazier :P

PR #2083, opened 2025-08-20
feat: mediaprojection improvements for android 14 and above

ArthurKun21:
## Pull Request Type

- [ ] Chore
- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

- #2140 
- #2129

## Description

Add new configuration for screen capture for android 15 and above

## Additional context

I will admit I haven't tested this one yet but might as well try

PR #2277, opened 2026-09-08
feat: mediaprojection capture whole screen as default

ArthurKun21:
- from #2083


reconman:
FGA works just fine when capturing only FGO, I don't see any reason to add this restriction

ArthurKun21:
iirc saw this problem with other devices, adding this will make it revert to previous behavior like from older android versions.

But if you think this still doesn't help then we can just close this PR.

reconman:
On my Android 16 device, it's not really "reverting to old behavior", but in reality it's disabling a dropdown value in the popup asking me what to record.

Disabling the single app option might make users suspect FGA is trying to spy on them.

PR #2282, opened 2026-09-12
feat(ui): add scrollbar indicator to drag and sort

ArthurKun21:


ArthurKun21:
ah old m3 version. makes a lot of sense

https://developer.android.com/develop/ui/compose/bom/bom-mapping

https://github.com/Fate-Grand-Automata/FGA/blob/e971d5d4f815576b5c7589b42338a19bb26dbecf/gradle/libs.versions.toml#L39

https://github.com/Fate-Grand-Automata/FGA/blob/e971d5d4f815576b5c7589b42338a19bb26dbecf/gradle/libs.versions.toml#L105

will revisit once the deps got updated

